PR #4, opened 2016-03-11
Adding simulation implementation for cordova-plugin-battery-status.

albertinadx:
Hi @TimBarham ! I've developed as part of this PR the simulation implementation of the cordova core plugin battery-status.
I checked how ripple does it, but I wanted to try and propose a different kind of simulation.
In this implementation, instead of letting the simulation to fire the `batterystatus`, `batterylow` and `batterycritical` events based on the changes from the simulation UI, is the actual navigator.battery instance who fires the event. To do that, the simulation implementation is based on the native logic of the plugin. On `Battery.start` exec call I save the success and error callbacks and use them to simulate the native call to `CallbackContext.sendPluginResult` and `CallbackContext.error`. So, the navigator.battery._start and navigator.battery._error callbacks are executed and the logic of firing the event based on the battery level and thresholds values defined, remains a responsibility of the plugin itself. If in an update the thresholds values are updated, we do not need to update the simulation code.

Thanks!

### DIFF
--- a/src/plugins/cordova-plugin-battery-status/battery.js
+++ b/src/plugins/cordova-plugin-battery-status/battery.js
@@ -1,0 +1,88 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+
+var db = require('db'),
+    constants = require('sim-constants');
+
+var _callbacks = {
+        sendPluginResult: null,
+        error: null
+    },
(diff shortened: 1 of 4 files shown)


albertinadx:
oops! wrong user!

PR #5, opened 2016-03-11
Adding simulation implementation for cordova-plugin-battery-status.

albertinad:
Hi @TimBarham ! I've developed as part of this PR the simulation implementation of the cordova core plugin battery-status.

I checked how ripple does it, but I wanted to try and propose a different kind of simulation.
In this implementation, instead of letting the simulation to fire the `batterystatus`, `batterylow` and `batterycritical` events based on the changes from the simulation UI, is the actual navigator.battery instance who fires the event. To do that, the simulation implementation is based on the native logic of the plugin. On Battery.start exec call I save the success and error callbacks and use them to simulate the native call to `CallbackContext.sendPluginResult` and `CallbackContext.error`. So, the `navigator.battery._start` and `navigator.battery._error` callbacks are executed and the logic of firing the event based on the battery level and thresholds values defined, remains a responsibility of the plugin itself. If in an update the thresholds values are updated, we do not need to update the simulation code.

Thanks!

### DIFF
--- a/src/plugins/cordova-plugin-battery-status/battery.js
+++ b/src/plugins/cordova-plugin-battery-status/battery.js
@@ -1,0 +1,91 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+
+var db = require('db'),
+    constants = require('sim-constants');
+
+var _callbacks = {
+        sendPluginResult: null,
+        error: null
+    },
+    _batteryInfo = {};
+
+var battery = {
+    listenerEnabled: false,
+
+    initialize: function (message) {
+        var isPlugged = db.retrieve(constants.BATTERY_STATUS.IS_PLUGGED_KEY);
+        _batteryInfo.level = db.retrieve(constants.BATTERY_STATUS.BATTERY_STATUS_KEY) || 100;
+        _batteryInfo.isPlugged =  (isPlugged) ? (isPlugged === 'true') : true;
+    },
+
+    /**
+     * Update the current battery level, save it and then notify using the callback passed on
+     * Battery.start call.
+     * @param {number} level
+     */
+    updateBatteryLevel: function (level) {
+        _batteryInfo.level = level;
+
+        db.save(constants.BATTERY_STATUS.BATTERY_STATUS_KEY, level);
+
+        notifyBatteryStatusChanged();
+    },
+
+    /**
+     * Update the current plugged state.
+     * @param {boolean} isPlugged
+     */
+    updatePluggedStatus: function (isPlugged) {
+        _batteryInfo.isPlugged = isPlugged;
+        db.save(constants.BATTERY_STATUS.IS_PLUGGED_KEY, isPlugged);
+
+        notifyBatteryStatusChanged();
+    },
+
+    /**
+     * The exec call to Battery.start executes this function.
+     * @param {function} success The function used to simulate the native call to CallbackContext.sendPluginResult.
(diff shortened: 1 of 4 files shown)


src/plugins/cordova-plugin-battery-status/sim-host.js
```diff
@@ -0,0 +1,40 @@
...
+    var battery = require('./battery');
+
+    battery.initialize();
+
+    function initialize() {
+        var labelBatteryLevel = document.getElementById('battery-level-label'),
+            levelRange        = document.getElementById('battery-level'),
+            pluggedCheckbox   = document.getElementById('is-plugged');
+
+        function updateBatterLevelText(level) {
```
TimBarham: Typo :smile: updateBatter**y**LevelText()

TimBarham:
Thanks @albertinad, this is great! I agree that this is the right approach (vs Ripple).

I only have one small comment - a typo.

src/plugins/cordova-plugin-battery-status/sim-host.js
albertinad: Oops! Fixed, thanks! I will be more careful next time.

src/plugins/cordova-plugin-battery-status/battery.js
```diff
@@ -0,0 +1,89 @@
...
+    },
+    _batteryInfo = {};
+
+var battery = {
+    listenerEnabled: false,
+
+    initialize: function (message) {
+        var isPlugged = db.retrieve(constants.BATTERY_STATUS.IS_PLUGGED_KEY);
+        _batteryInfo.level = db.retrieve(constants.BATTERY_STATUS.BATTERY_STATUS_KEY) || 100;
+        _batteryInfo.isPlugged =  (isPlugged) ? (isPlugged === 'true') : true;
```
albertinad: @TimBarham  I found an issue here that I've fixed. Boolean values are retrieved as string from db, so I've modified this in order to save the boolean value of isPlugged, otherwise the fallback is true.
When the saved value was false, in the previous implementation, the isPlugged value ends up being true, that was wrong.

albertinad:
Hi @TimBarham , do you have any updates on this??

albertinad:
Great! sure, thank you @guillaumejenkins!

src/plugins/cordova-plugin-battery-status/battery.js
```diff
@@ -0,0 +1,89 @@
...
+        notifyBatteryStatusChanged();
+    },
+
+    /**
+     * Update the current plugged state.
+     * @param {boolean} isPlugged
+     */
+    updatePluggedStatus: function (isPlugged) {
+        _batteryInfo.isPlugged = isPlugged;
+        db.save(constants.BATTERY_STATUS.IS_PLUGGED_KEY, isPlugged);
```
dlebu: Do we need a `notifyBatteryStatusChanged()` call here as well?

dlebu:
This looks good to me. I only had one minor question. Feel free to merge it when you are ready!

src/plugins/cordova-plugin-battery-status/battery.js
albertinad: Good catch! yes, we need to notify here also. I will make that change.

albertinad:
I've squashed my commits into a single one since this branch is old.
Thanks!

PR #6, opened 2016-03-11
Supporting simulation for cordova-plugin-network-information.

albertinad:
Hi @TimBarham ! I've implemented the simulation for the network-information plugin.
I'm running the automated test suite of the cordova mobilespec app, with taco-simulate, and it the results are really good, I'm implementing the support of the missing plugins based on the automated test suite results and manual test using the mobilespec app.

platform: iOS
mobilespec-network-plugin

Thanks!

### DIFF
--- a/src/platforms/windows/app-host-clobbers.js
+++ b/src/platforms/windows/app-host-clobbers.js
@@ -1,5 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 
+/*
+ * Windows API definition taken from:
+ * - https://msdn.microsoft.com/library/windows/apps/windows.networking.connectivity.networkconnectivitylevel.aspx
+ */
+
 module.exports = {
     // Сlobbing WinJS.Application.addEventListener
     // for proper cordova-plugin-test-framework initialization
@@ -7,6 +12,18 @@ module.exports = {
     WinJS: {
         Application: {
             addEventListener: function () {}
+        }
+    },
+    Windows: {
+        Networking: {
+            Connectivity: {
+                NetworkConnectivityLevel: {
+                    none: 0,
+                    localAccess: 1,
+                    constrainedInternetAccess: 2,
+                    internetAccess: 3
+                }
+            }
         }
     }
 };
--- a/src/plugins/cordova-plugin-network-information/app-host-clobbers.js
+++ b/src/plugins/cordova-plugin-network-information/app-host-clobbers.js
@@ -1,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+
+module.exports = {
+    Windows: {
+        Networking: {
+            Connectivity: {
+                NetworkInformation: {
+                    getInternetConnectionProfile: function () {
+                        return {
+                            getNetworkConnectivityLevel: function () {
+                                // defined in platforms/windows/app-host-clobbers.js
+                                return Windows.Networking.Connectivity.NetworkConnectivityLevel.internetAccess;
+                            },
+                            networkAdapter: {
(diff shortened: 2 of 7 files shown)


albertinad:
Hi @TimBarham , do you have any updates on this??

src/plugins/cordova-plugin-network-information/app-host-clobbers.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
albertinad: This clobbers are needed in order to be able to load the script `src/windows/NetworkInfoProxy.js` when running the simulation for `windows` platform.

src/platforms/windows/app-host-clobbers.js
```diff
@@ -8,5 +13,17 @@ module.exports = {
         Application: {
             addEventListener: function () {}
         }
+    },
+    Windows: {
+        Networking: {
+            Connectivity: {
+                NetworkConnectivityLevel: {
```
albertinad: This is needed too by the network info plugin when loading `src/windows/NetworkInfoProxy.js` for `windows` platform. The reason I've added it here as a common API is because it is an enumeration defined that any other plugin could use. Since it is well-defined, there's no need that any simulation plugin re-define it again only for the plugin. This can be shared.

src/plugins/cordova-plugin-network-information/network.js
```diff
@@ -0,0 +1,53 @@
...
+        this.connectionType = db.retrieve(simConstants.NETWORK_TYPE_KEY) || this._defaultConnectionType;
+    },
+
+    /**
+     * @param {string} type
+     */
+    updateNetworkType: function (type) {
+        this.connectionType = (this.ConnectionTypes[type]) ? type : 'UNKNOWN';
+
+        db.save(simConstants.NETWORK_TYPE_KEY, this.connectionType);
```
dlebu: I wonder if saving the `ConnectionTypes` values instead of property names in the db would simplify the code. That way, there would not be a need to know the `ConnectionTypes` property names (and use them as string constants), but use the values directly. 

The default connection type would change to:
```javascript
_defaultConnectionType: this.ConnectionTypes.CELL_4G;
```

The save call would change to:
```javascript
this.connectionType = type;
```
and the `sendUpdate` function would just send over the value - we're the only ones calling `updateNetworkType` anyway.

dlebu:
Looks good to me! The suggestion I had is totally optional.

src/plugins/cordova-plugin-network-information/network.js
albertinad: The code could be simpler, but by saving the values instead of the property, we are loosing the opportunity in the future to add localize the values for the current locale, or display more information. In the case of WiFi there's nothing to localize, but with the UNKNOWN yes. Today the value and the key are the same, but instead of CELL_4G we can display 4G (LTE) or something similar.

albertinad: I read again your comment, you're right, ignore my previous comment! I will make the change you are suggesting. Thanks!

albertinad:
I've updated the implementation following your suggestion, now we are saving the connection value to the db, as well as displaying the connection value in the list, just to reflect the real connections values.

screen shot 2016-05-06 at 1 32 56 pm

dlebu:
@albertinad Thanks for making this change. Again, this looks good to me - feel free to merge it to master.

albertinad:
Thanks @dlebu ! I just added one more change, an update to the README file to add the 2 new core plugins that we provide simulation: battery-status and network-information.
Merging!

PR #7, opened 2016-03-20
Updating geolocation and globalization plugins simulation support

albertinad:
Hi @TimBarham , this PR is for updating the simulation support for geolocation and globalization support:
geolocation:
- Updating `PositionError` object to include the error constants as part of the object instance (based on the plugin definition)

globalization:
- Completing not supported functions by using accounting.js library
- Returning formatting pattern when it is available
- Updating `getFirstDayOfTheWeek` to return the index of the first day instead of the name, by following the plugin definition

For testing these changes, I've used the cordova mobilespec app with only the tests that are related to the geolocation and globalization plugins, see the results below:

Before these changes:

tests-before
*Specs: 56 - failures: 14*

After these changes:

tests-after
*Specs: 56 - 0 failures*

Thanks!


dlebu:
This looks good to me! Awesome test results.

albertinad:
Thanks! merging.

PR #8, opened 2016-04-11
Simulation support for cordova-plugin-device-orientation

albertinad:
Hi @TimBarham! This PR is for adding simulation support for the cordova-plugin-device-orientation, the device compass. I've followed the implementation of the device accelerometer and try to achieve something similiar: a widget similar to a device compass and actions to update the heading:

device-orientation-widget

Since the geolocation plugin, and API, add an API to get the current compass heading, I've modified the implementation of `Messages`: emit, on and off functions, in order to support emitting and listening for global events that are out of the plugin scope. So, geolocation and device-orientation emits a global event when the heading is updated, and both listen for the same event. When the geolocation plugin is present, and orientation too, then both are synchronized when the heading value is updated.

To add the simulation support, I've used the mobilespec app and run the suite related to device-orientation:

Before:
device-orientation-001

After:
device-orientation-002

I wanted to share this PR as soon as possible to get feedback :), in the meantime, I will improve the documentation.

Thanks!

cc @busykai


src/plugins/cordova-plugin-device-orientation/compass-widget.js
```diff
@@ -0,0 +1,304 @@
...
+ * Get the direction according to the heading value.
+ * @param {number} heading A number from 0 to 359.99.
+ * @return {string} direction It can be one of the following: N, NE, E, SE, S, SW, W or NW.
+ */
+CompassWidget.getDirection = function (heading) {
+    if (heading >= 337.5 || (heading >= 0 && heading <= 22.5)) {
+        return CompassWidget.Direction.N;
+    }
+
+    if (heading >= 22.5 && heading <= 67.5) {
```
dlebu: To avoid intersection of the intervals, I would use the equal comparator only at one side of the comparison like this:
```
if (heading > 22.5 && heading <= 67.5) {
```

or 

```
if (heading >= 22.5 && heading < 67.5) {
```

Same for the other intervals.

src/plugins/cordova-plugin-device-orientation/compass.js
```diff
@@ -0,0 +1,54 @@
...
+        if (typeof success === 'function') {
+            success(compass.getHeadingResult());
+        }
+    },
+
+    /**
+     * The exec call to Compass.stopHeading executes this function.
+     */
+    stopHeading: function () {
+        // TODO implement for iOS
```
dlebu: Can you open an issue for this so that we don't forget about it?

dlebu:
Minor comments. This looks good to me!

src/plugins/cordova-plugin-device-orientation/compass-widget.js
albertinad: Good catch! I will make the change as part of the next commit.

src/plugins/cordova-plugin-device-orientation/compass.js
albertinad: After digging a little into the implementation of the cordova-device-orientation plugin, there's no need to have some implementation specific for iOS simulation. The reason of the TODO comment was that for iOS there's no need to create timers in javascript side, comparing for the other platforms. This is identified by a UUID value generated using utils module from cordova. But when running on the browser and simulator, that validation never applies, and the timers are set in place.
Removing this TODO and executing the success callback.

src/plugins/cordova-plugin-device-orientation/compass-widget.js
```diff
@@ -0,0 +1,304 @@
...
+    W: 'W',
+    NW: 'NW'
+};
+
+/**
+ * Get the direction according to the heading value.
+ * @param {number} heading A number from 0 to 359.99.
+ * @return {string} direction It can be one of the following: N, NE, E, SE, S, SW, W or NW.
+ */
+CompassWidget.getDirection = function (heading) {
```
albertinad: I don't like to have this function duplicated in both plugins: device-orientation and geolocation.
Both plugins needs to be synchronized about the heading value (orientation), so when these ranges needs to be updated, we need to change it for both simulation plugins.
What do you think about adding this function to a common module, like an utility, that is shared between all plugins and provided by cordova-simulate? By provided by cordova-simulate I mean that, in a future it could be really good to see this simulation logic implemented as part of the plugin itself and in the plugin repository, but the utilities modules provided by cordova simulate.
If we follow that approach, we share the same ranges to determined the heading value.

albertinad: This is the current [utility](https://github.com/Microsoft/cordova-simulate/blob/master/src/modules/utils.js) module that we provided for sim-host. For the type of utilities function it provides, maybe is not the best place for this function, but we can add a new one named like `navigation-utils.js`, what do you think?

dlebu: I think that is reasonable. I agree that the declaring the ranges in one place only is a good idea.

albertinad: Great! I will make that change then.

albertinad:
@dlebu I've updated this PR with the following changes:
- The common function to get the heading direction is now implemented in the common module utils. I find a navigationHelper function in there that expose all utility related to navigation, so it wasn't needed to add a new module, just updating the current.
- Sending telemetry events for the UI usages: input field and compass widget.
- Removing unused function and variables from geolocation sim-host.js module.

What do you think guys about this? If everything is ok I will merge it.
Thanks!

dlebu:
LGTM! Thanks @albertinad!

albertinad:
Merged! thanks!

PR #11, opened 2016-04-29
Merge taco-simulate and taco-simulate-server

dlebu:
Submitting an initial pull request for review. Summary:

- Moved `taco-simulate` files under `src/sim-host`
- Added a new `simhostroot` parameter for 3rd party simulation host UI. Defaults to `./src/sim-host`.
- Updated names, URLs in `package.json`
- Bumped the version to `0.2.0`

One thing I did not port over since I wanted to get feedback first. It is the [server file in taco-simulate](https://github.com/Microsoft/taco-simulate/blob/master/src/server/server.js). I wonder if we could solve this browser specific css matter directly in the `.css` file, or in the sim-host `.js`, without relying on knowledge about the server. Any suggestions are appreciated here.

@TimBarham @albertinad @guillaumejenkins @busykai @hamiltonia


albertinad:
@dlebu I agree. We should solve this css specific in sim-host side, not in the server side. We can detect the browser and load different css, or, when the browser is identified, add some id to the body, like `#no-chrome` (probably a better name here) and then, the css rules
{code}
body /deep/ button:active,
#no-chrome body button:active {
    /* here goes the rules */
}
{code}

src/sim-host/socket.js
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
albertinad: This file is not just a UI related module. It is part of the sim-host client side implementation but it is critical for the main communication between app-host and server. Is the sim-host side implementation of the protocol for simulation.
If we enable developers to change this, they must know the protocol, and maybe they end up doing the same than here, when they only wants to provide a customised UI.
So, maybe we can do this:
- provide the default implementation but from the server side that is only for sim-host, so we should move `socket.js` to some folder in the server
- Let them load it on `sim-host.js` module, by default our implementation will load the socket implementation from our server, instad of doing this https://github.com/Microsoft/cordova-simulate/pull/11/files#diff-c1aa19020ad6303656f2fbf0b9d79df4R6, we can do `var socket = require('socket')`. We can also rename the script name to be more representative, since it handles the sim-host side of the protocol, it could be protocol.js, under some sim-host folder in the server.
- if the developers of the new sim-host UI needs to tweak a little the protocol for them, wait for something in the UI before connecting or something, then they can provide this file and must follow the protocol defined for the communication, then they should do `var protocol = require('my-modified-protocol')` instead of the default one.

I have my doubts about the last point: let developers modified the sim-host side protocol. If modifying the protocol is a requirement for other developers, maybe the recommendation for them is to clone the whole repo and make the modification as needed in the protocol. What do they need that they provide a new protocol? If it is refreshing sim-host side in a different way than just calling `document.location.reload(true);`, or handling the failures differently, by not logging the error but doing something else, then we can provide some optional configuration to the initialization call that enable developers to customize some actions.

guillaumejenkins: What about moving all the UI-related code to a `src/sim-host/UI/` folder, and have the consumers that want a custom UI pass in the path to their `UI/` folder instead of the `sim-host/` root? This way, we can keep all the sim-host files under `src/sim-host/` in the repo, while still making it clear that core files like this one should not be modified.

This of course assumes that we don't want to support custom protocols. I think it's perfectly fine to not allow that. Like you said, if they need such big tweaks, then it will be better for them to fork the repo.

Whatever we do, I would really prefer having everything related to sim-host be under a single `src/sim-host/` directory.

albertinad: That works too! It is even more clear to separate what is UI-related and what is not, and pass the path to the UI and not everything that can impact too to the protocol.
Thanks!

busykai: just let's use lowercase names for directories. :)

dlebu: I created two new directories: `ui` and `protocol`. `socket.js` went in `protocol`, while all the user interface files went into `ui`.

package.json
```diff
@@ -1,18 +1,17 @@
 {
-  "name": "taco-simulate-server",
-  "version": "0.1.3",
-  "description": "Server and app host module for taco-simulate (Cordova plugin simulation).",
-  "main": "src/server.js",
+  "name": "cordova-simulate",
+  "version": "0.2.0",
```
TimBarham: Generally don't bump the version until we actually do a release. Otherwise people who install from the repo will be at the same version as the release when it finally comes out. Either leave this as `0.1.3` until we release, or do what we do with Cordova and make it something like `0.2.0-dev` (which is treated as a lower version than `0.2.0`) until we release.

src/server/log.js
```diff
@@ -14,7 +14,6 @@ function error(error) {
     }
     console.log(chalk.red.bold('SIM ERROR: ' + error));
     config.server && config.server.close();
-    process.exit(1);
```
TimBarham: Why did this need to be removed?

src/sim-host/ui/sim-host.css
```diff
@@ -0,0 +1,507 @@
...
+    position: absolute;
+    top: 5px;
+    right: 10px;
+}
+
+.cordova-main {
+    padding-bottom: 40px;
+}
+
+/* Chrome 29+ */
```
TimBarham: I wouldn't do this. It's going to be too much of a pain to maintain two separate sets of virtually identical CSS (which is why I didn't do it this way in the first place). The existing approach isn't ideal, but ultimately we will change how we render so we don't require such differences (/deep/ is deprecated anyway) - better to keep maintenance simple in the meantime.

src/simulate.js
```diff
@@ -53,7 +52,37 @@ module.exports = function (opts, simHostOpts) {
...
-function parseStartPage(projectRoot) {
+var closeServer = function() {
+    return server.server && server.server.close();
+};
+
+var launchBrowser = function(target, url) {
+    return cordovaServe.launchBrowser({ target: target, url: url });
+};
+
+var simulate = function(opts) {
```
TimBarham: How do you plan to support alternative heads (UI renderers etc)?

README.md
```diff
@@ -23,7 +23,7 @@ Where opts is an object with the following properties (all optional):
 * **target** - the name of the browser to launch your app in. Can be any of the following: `chrome`, `chromium`, `edge`, `firefox`, `ie`, `opera`, `safari`. Defaults to `chrome`.
 * **port** - the desired port for the server to use. Defaults to `8000`.
 * **dir** - the directory to launch from (where it should look for a Cordova project). Defaults to cwd.
-* **simhostroot** - the directory containing the UI specific files of the simulation host. Defaults to the bundled simulation host files, found in `src/sim-host`.
+* **simhostui** - the directory containing the UI specific files of the simulation host. Defaults to the bundled simulation host files, found in `src/sim-host`.
```
guillaumejenkins: I would also update the description: "... Defaults to the bundled simulation host **UI** files, found in src/sim-host**/ui**."

README.md
```diff
@@ -1,7 +1,228 @@
...
+simulate(opts);
+```
+
+Where opts is an object with the following properties (all optional):
+
+* **platform** - any Cordova platform that has been added to your project. Defaults to `browser`.
+* **target** - the name of the browser to launch your app in. Can be any of the following: `chrome`, `chromium`, `edge`, `firefox`, `ie`, `opera`, `safari`. Defaults to `chrome`.
+* **port** - the desired port for the server to use. Defaults to `8000`.
+* **dir** - the directory to launch from (where it should look for a Cordova project). Defaults to cwd.
+* **simhostui** - the directory containing the UI specific files of the simulation host. Defaults to the bundled simulation host files, found in `src/sim-host`.
```
guillaumejenkins: My previous comment got hidden because of "commenting on a previous diff", so I'm adding it here:

I would also update the description: "... Defaults to the bundled simulation host UI files, found in src/sim-host/ui."

src/server/log.js
guillaumejenkins: Exiting makes sense for standalone taco-simulate on the command-line, but programmatic consumers might prefer to not have taco-simulate kill them unexpectedly and instead deal with any errors themselves.

@dlebu should this at least re-throw the error so that programmatic consumers be aware that a fatal error occurred? You left the call to `server.close()`, which means the whole thing will be in a bad state.

package.json
dlebu: Changing to `0.2.0-dev`.

src/server/log.js
dlebu: I am removing the `server.close` call as well from `log.js`. I added them to `bin/simulate` for command line consumption scenarios. As @guillaumejenkins said, programmatic consumers should handle errors directly.

src/sim-host/ui/sim-host.css
dlebu: I agree that a subset of the CSS rules are duplicated and that is not ideal. However, from an architecture perspective, I believe this approach offers a better separation between the simulation UI and the server. If the only concern is duplication, I think we should be fine.

What do you mean when you say we will change how we render? My understanding was that out of all the approaches investigated, this was the one offering the most flexibility.

src/simulate.js
dlebu: I added a [new option, simhostui](https://github.com/Microsoft/cordova-simulate/pull/11/files/334e563898a5d13a539f575dcbebac6f4d6e711f#diff-6ddd72d4edcfc54d2078a3ddee13c0c7R19) - which currently supports an arbitrary UI directory, very similar to how the old `taco-simulate` and `taco-simulate-server` integration used to work. Please let me know if this answers your question. If we need to support more, or I missed something while porting over the changes, let's talk about it.

README.md
dlebu: Fixed, thanks!

src/sim-host/ui/sim-host.css
TimBarham: "better separation between the simulation UI and the server" - since any UI implementation can insert its own server middleware, this isn't really an issue.

What we have currently is a simple and temporary hack which, while being a hack, is easy to work with and maintain. The problem is that switching this to have a bunch of duplicated CSS doesn't improve anything, while make maintenance more difficult.

"What do you mean when you say we will change how we render?" - before long we're going to have to do something different, at least with the CSS, since the /deep/ combinator has been deprecated. It may well be supported for some time, but its use will result in warning messages output to the Chrome console, which is something that we should aim to avoid.

TimBarham:
Regarding @albertinad's suggestion with the CSS - I already tried something like that and it doesn't work. The `/deep/` combinator is treated as a syntax error by other browsers, and the entire rule is ignore. Believe me, at the time I implemented this I tried a bunch of different approaches, and what we have now was the least evil :). Particularly for something that is going to need to be temporary.

Regarding separation of sim-host and server: I completely agree that however this is handled, it should of course be the responsibility of the sim-host (that is, the UI renderer). Which it is, in the current implementation. Any sim-host can inject middleware into the server for situations just such as this one.

While the current approach isn't ideal, it is the one that allows the easiest maintenance going forward until we come up with some alternative to the `/deep/` combinator.

src/simulate.js
```diff
@@ -70,14 +99,12 @@ function parseStartPage(projectRoot) {
...
 module.exports.dirs = require('./server/dirs');
 module.exports.app = server.app;
-module.exports.log = log;
-
-Object.defineProperty(module.exports, 'server', {
-    get: function () {
-        return config.server;
-    }
-});
+module.exports.log = log;
```
albertinad: This file seems to miss a new line at the end.

src/server/log.js
TimBarham: Oh, right - that code was from when this was CLI only. Regarding removing server.close - who will handle closing the server in the API scenario?

src/simulate.js
dlebu: Fixing...

src/sim-host/ui/sim-host.css
albertinad: I agree with @TimBarham on this, duplicating styles will make the maintenance and improving the sim host UI harder than keeping the express middleware. I believe that we can leave with that middleware for a while, knowing that is a hack and that in the future (not so far) will be removed.
If it happens that we need to keep the `deep` combinator for Chrome, then we can think of other approaches.

albertinad:
@TimBarham I didn't try my suggestion, I didn't know that it doesn't know, thanks! I think that this is not a big issue that we can live with for a while. There might be other approaches too, like instead of processing the css file before sending it (middleware work), we can do that by running a kind of gulp task that generates the css needed for browsers: chrome and others, and have some logic to detect the browser on sim-host and load one or another css. Once we make changes to the style and we want to  What we can save here is duplication of styles, since we are going to make the changes on the main css (the template), and we avoid to use a server middleware to help the client to render the UI. I don't think that what I describe is a good approach, or better, it is just another idea that will still add complexity for developing and distributing cordova-simulate. But could help.
It is true that the approach that we have today is easiest for maintenance and if it is temporary, we can live with it :)

dlebu:
Alright guys, majority wins! I am reverting the change and putting in the middleware back in business. :)

src/server/log.js
dlebu: I added [a function for that](https://github.com/Microsoft/cordova-simulate/pull/11/files/334e563898a5d13a539f575dcbebac6f4d6e711f#diff-6ddd72d4edcfc54d2078a3ddee13c0c7R55).

src/simulate.js
```diff
@@ -12,14 +12,13 @@ var Q = require('q'),
...
-        simHostUrl;
-
+var launchServer = function(opts) {
     opts = opts || {};
-
+    
     var platform = opts.platform || 'browser';
-    var target = opts.target || 'chrome';
+    var simHostRoot = opts.simhostui || path.join(__dirname, 'sim-host', 'ui');
+    var simHostOpts = { simHostRoot: simHostRoot };
```
albertinad: These 2 lines can be combined in a single line:
```
var simHostOpts = {
    simHostRoot: opts.simhostui || path.join(__dirname, 'sim-host', 'ui')
};
```

src/simulate.js
```diff
@@ -12,14 +12,13 @@ var Q = require('q'),
...
+    var appUrl, simHostUrl;
 
     config.platform = platform;
     config.simHostOptions = simHostOpts || {};
```
albertinad: This validation is not needed since `simHostOptions` will always be defined.

src/simulate.js
TimBarham: Should validate `simhostui`.

README.md
```diff
@@ -1,7 +1,228 @@
...
+The contents of the `cordova-panel` element can be regular HTML, or the various custom elements which are supported
+(see the existing plugin files for more details).
+ 
+This file shouldn't contain any JavaScript (including inline event handlers), nor should it link any JavaScript files.
+Any JavaScript required can be provided in the standard JavaScript files described below, or in additional JavaScript
+files that can be included using `require()`.
+ 
+*sim-host-dialogs.html*
+
+This defines any dialogs that will be used (dialogs are simple modal popups � such as used for the Camera plugin). At
```
TimBarham: Something got screwed with the readme in its most recent update. Could you change each � back into a dash?

package.json
TimBarham: Sounds good. Going forward, this will mean each time we do a release we remove the `-dev` (and either going with this version or bump it if the changes warrant that), then immediately following the release we bump the patch version and add the `-dev` again.

src/server/log.js
TimBarham: Ah, I missed that. But shouldn't the server still be responsible for closing itself in this scenario, rather than expecting the API consumer to do that?

src/simulate.js
```diff
@@ -70,14 +99,12 @@ function parseStartPage(projectRoot) {
     }
 
     return 'index.html'; // default uri
-}
+};
 
+module.exports = simulate;
+module.exports.launchBrowser = launchBrowser;
+module.exports.launchServer = launchServer;
+module.exports.closeServer = closeServer;
```
TimBarham: Is there still a need to expose any of these externally? Previously `launchBrowser` and `launchServer` were exposed so taco-simulate could call them.

src/simulate.js
```diff
@@ -53,7 +52,37 @@ module.exports = function (opts, simHostOpts) {
...
+var launchBrowser = function(target, url) {
+    return cordovaServe.launchBrowser({ target: target, url: url });
+};
+
+var simulate = function(opts) {
+    
+    var target = opts.target || 'chrome';
+    var simHostUrl;
+
+    return launchServer(opts)
```
TimBarham: Do we still need `launchServer()` separate from `simulate()`? Or, if we do keep them separate (which I'm fine with), we don't need to have the `catch` handler in both places.

README.md
dlebu: Fixing...

src/simulate.js
dlebu: Removing...

src/simulate.js
dlebu: I agree.

src/simulate.js
dlebu: We do need both. In VSCode, we only launch the server and the app-host via `launchServer()` and `launchBrowser`. Sim-host is launched differently.

dlebu: I removed the handler from `launchServer`.

dlebu:
@TimBarham @albertinad @guillaumejenkins I addressed all the comments - please let me know if this looks good to you - I'd like to merge it in as soon as possible.

src/sim-host/ui/sim-host.css
dlebu: The middleware is back in.

albertinad:
Looks good to me! :)

dlebu:
Thanks everyone, merging to master.

I looked at @TimBarham's suggested [guide for rebasing](https://www.atlassian.com/git/tutorials/merging-vs-rebasing/). The Golden Rule of Rebasing states that it should never be used on public branches, but people in our team already took a dependency on this branch. With that in mind, I will merge this PR directly. I think the rebase option is nice to have, but it can get in our way and be error-prone if not done properly.

TimBarham:
The guide for rebasing wasn't really talking about temporary feature branches. I would say people take dependencies on in-progress feature branches at their own risk. At the very least before you merge a change like this into master you should squash your commits down into something more useful (ultimately what we want is the clearest history for someone down the track to understand the change, and that is rarely the same as the raw commits that occur as the PR progresses). Since that will require rewriting the history of the feature branch anyway, there is no additional problem rebasing.

An alternative approach, which leaves the original branch intact, is to do the squash and rebase in a different branch (the github PR won't close automatically, but you can add "This closes #xx" to the final commit to associate the commit with the PR and close the PR).

PR #12, opened 2016-05-05
Using a common third-party libraries in sim-host.

albertinad:
Hi folks, after the merge of the server and sim-host, we have 2 third-party folders and a duplicated library: socket io client.
The motivation of this PR is to propose an alternative solution for this issue. Today, when we start the server and prepare the routes, we serve statically a third-party folder, and that is used only by app-host. On the other hand, sim-host UI specify its own third-party dependencies that are being served too. As a solution, we can combined both third-party folder and share between app-host and sim-host the third-party libraries served. 
So, by default, the sim host UI implementation that we provide do not provide third-party libraries, it just use the ones served by the server.

Thanks!


src/sim-host/ui/sim-host.html
```diff
@@ -6,8 +6,8 @@
     <meta charset="UTF-8">
     <title></title>
     <link href="sim-host.css" rel="stylesheet"/>
-    <script src="third-party/webcomponentsjs/webcomponents.min.js"></script>
-    <script src="third-party/socket.io/socket.io-1.2.0.js"></script>
+    <script src="/simulator/thirdpartywebcomponents.min.js"></script>
```
dlebu: Missing a slash here:
`/simulator/thirdparty/webcomponents.min.js`

dlebu:
Although it makes sense to reuse common libraries, this change will couple the server and sim-host default UI a little bit more. Given the `.css` hack earlier this week, I am fine with this change, but I am opening the discussion in case anybody has a different opinion.

src/sim-host/ui/sim-host.html
albertinad: oops! thanks, I will fix it.

albertinad:
@dlebu I see your point. The default implementation is a little more coupled with the server with this approach, but it is optional, others sim-host can provide their own third-party libraries. I'm happy to open the discussion, that is the point of this PR :)

dlebu:
Awesome. I think we have critical majority, so feel free to merge this one as well @albertinad .

albertinad:
Great, thanks!

PR #13, opened 2016-05-05
Added live reload feature, and added many telemetry points

guillaumejenkins:
- Added live reload feature, which is enabled by default. CSS changes are propagated without a page reload; JS and HTML changes trigger a full page reload.
- Improved the existing telemetry infrastructure, and added many telemetry points


src/app-host/live-reload-client.js
```diff
@@ -0,0 +1,146 @@
...
+    function reloadPage() {
+        window.location.reload(true);
+    }
+
+    /**
+     * Fetch all the local stylesheets from the page.
+     *
+     * @returns {Object} The list of local stylesheets keyed by their relative path.
+     */
+    function getLocalStylesheets() {
```
dlebu: What is the livereload behavior for internal (`<style>` tag in HTML `<head>` tag) and inline CSS?

guillaumejenkins: Changing those will trigger a file system event on the HTML file, which means we will detect it the same as any other HTML change. A full reload will happen.

src/app-host/app-host.js
```diff
@@ -72,22 +75,27 @@ socket.on('exec-failure', function (data) {
     }
 });
 
+socket.on('start-live-reload', function () {
+    livereload.start(socket);
+});
+
+telemetry.init(socket);
```
dlebu: Do we need to call `init()` every time? What if we don't want to enable telemetry?

src/plugins/cordova-plugin-camera/sim-host.js
```diff
@@ -1,6 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 
 var dialog = require('dialog');
+var telemetry = require('../../client-common/client-telemetry-helper');
+
+var baseProps = {
+    plugin: 'cordova-plugin-camera',
```
dlebu: Nit: Can we use the folder name here instead of hardcoding the plugin name?

src/plugins/cordova-plugin-device-motion/sim-host.js
```diff
@@ -197,47 +208,58 @@ function createCanvas() {
...
-            _offsets.x = (_offsets.x + _oldX - e.offsetX) % 360;
-            _offsets.y = (_offsets.y + _oldY - e.offsetY) % 360;
-
-            _alpha = _alpha || 0;
-
-            // enforce gamma in [-90,90] as per w3c spec
-            _gamma = -_offsets.x;
-            if (_gamma < -90) {
-                _gamma = -90;
+        if (_mouseDown) {
```
dlebu: We should make sure we are not breaking any functionality with these changes. I wonder if we can run the mobile spec tests before and after the change and compare the results. @albertinad, are the tests available for all the supported plugins so far? Can you point us to some documentation on how to run them?

src/plugins/cordova-plugin-device/sim-host.js
```diff
@@ -298,6 +305,7 @@ function initialize() {
     deviceList.onchange = handleSelectDevice;
     deviceList.value = 'WVGA';
     handleSelectDevice();
+    registerTelemetryEvenst();
```
dlebu: Typo

src/plugins/cordova-plugin-camera/sim-host.js
guillaumejenkins: I initially thought of doing something like that, but all plugin code gets bundled into sim-host.js under the simulation folder in the app, so we can't use things like `__dirname`. Essentially, we want to programmatically get the "design time" folder name, when the "run time" is different. I don't see a way of doing this without hardcoding something (if we don'T hardcode the plugin name, we'll need to hardcode the path to its folder, so we don't really gain anything).

src/plugins/cordova-plugin-device-motion/sim-host.js
guillaumejenkins: All I did was refactor the common `_mouseDown` into a parent `if`; the code before and after are semantically identical. I don't mind running the tests though, it can never hurt.

src/app-host/app-host.js
guillaumejenkins: Good idea, I will change it to call init when the server requests to, so telemetry never gets initialized if it is disabled.

src/plugins/cordova-plugin-geolocation/sim-host.js
```diff
@@ -77,6 +88,49 @@ module.exports = function(messages) {
...
+            --pendingMouseEvents;
+
+            if (pendingMouseEvents === 0) {
+                telemetry.sendUITelemetry(Object.assign({}, baseProps, { control: 'geo-map-container' }));
+            }
+        }, mouseEventHoldDelay);
+    }
+
+    function registerTelemetryEvents() {
+        document.getElementById('geo-latitude').onchange = telemetry.sendUITelemetry.bind(this, Object.assign({}, baseProps, { control: 'geo-latitude' }));
```
dlebu: Can we create a separate function that takes the element id as a parameter and attaches the `onchange` handler? We could then reuse it for all the element ids.

src/server/live-reload/live-reload-server.js
```diff
@@ -0,0 +1,34 @@
...
+    socket.emit('lr-full-reload');
+}
+
+module.exports.init = function (sock) {
+    if (!watcher) {
+        watcher = new Watcher(cssFileChanged, nonCssFileChanged);
+        watcher.startWatching();
+    }
+
+    socket = sock;
```
dlebu: Will this work for multiple app browsers at a time? We don't need to support it for right away, but @busykai mentioned he will be looking into supporting that, so it is good to keep that scenario in mind.

src/server/live-reload/watcher.js
```diff
@@ -0,0 +1,153 @@
...
+
+    if (!fileRelativePath) {
+        // fs.watch() doesn't always set the fileRelativePath argument properly. If that happens, let the user know.
+        log.warning('Could not reload the modified file because fs.watch() didn\'t specify which file was changed');
+
+        return;
+    }
+
+    // Make sure the modified file is either JS, HTML or CSS.
+    if (!fileIsOfType(fileRelativePath, ['.html', '.css', '.js'])) {
```
dlebu: Do we care about other resources, like images, referenced in html?

src/server/live-reload/watcher.js
```diff
@@ -0,0 +1,153 @@
...
+        // No need to prepare, because the server already does when app-host does a full reload.
+        preparePromise = Q.resolve();
+    }
+
+    // Invoke the appropriate callback.
+    preparePromise.then(function () {
+        if (isCss) {
+            this.cssChangedCallback(fileRelativePath);
+        } else {
+            this.nonCssChangedCallback(fileRelativePath);
```
dlebu: We should validate `this.cssChangedCallback` and `this.nonCssChangedCallback` to make sure they are defined and of function type.

src/plugins/cordova-plugin-device/sim-host.js
guillaumejenkins: Will be fixed in my next commit, thanks

src/plugins/cordova-plugin-geolocation/sim-host.js
guillaumejenkins: Will be fixed in my next commit

src/server/plugins.js
```diff
@@ -3,13 +3,19 @@
...
+    telemetry = require('./telemetry-helper');
 
 var pluginSimulationFiles = require('./plugin-files');
 
 var plugins = {};
+var pluginsTelemetry = {
+    simulatedBuiltIn: [],
+    simulatedNonBuiltIn: [],
+    notSimulated: []
+};
```
dlebu: I would extract this init code (for `plugins` and `pluginsTelemetry` variables) to a function and reuse it in `initPlugins()`. It feels like we are initializing there again.

src/simulate.js
```diff
@@ -110,4 +110,3 @@ module.exports.launchServer = launchServer;
 module.exports.closeServer = closeServer;
 module.exports.dirs = dirs;
 module.exports.app = server.app;
-module.exports.log = log;
```
dlebu: Is removing the '.log' export intentional? We should probably keep it for backward compatibility with consumers.

src/server/plugins.js
guillaumejenkins: Will be fixed in my next commit

src/simulate.js
guillaumejenkins: I think that's reasonable, will be added back in my next commit

src/server/live-reload/watcher.js
guillaumejenkins: Fixed in my next commit

src/server/live-reload/live-reload-server.js
guillaumejenkins: This is a good point. I think for now, however, it might not make much sense to support multiple browsers when cordova-simulate itself doesn't? I would prefer tracking this in an issue and making the changes when the multiple browser support PR is opened, what do you think?

dlebu: I think that is a good plan.

src/plugins/cordova-plugin-camera/sim-host.js
dlebu: sounds good - this is fine

src/server/live-reload/live-reload-server.js
busykai: @guillaumejenkins, @dlebu I think for now assuming 1-to-1 is fine.

src/server/live-reload/watcher.js
guillaumejenkins: I've changed the way live reload works. We no longer enforce specific file types (we listen for all changes). Also, for "resources", i.e. anything that is referenced in the html via a url, href or src attribute, we support refreshing without reloading the page. This includes but isn't limited to CSS and images.

src/server/live-reload/live-reload-server.js
```diff
@@ -1,26 +1,97 @@
...
+var refreshFileEventName = 'lr-refresh-file';
+var fullReloadEventName = 'lr-full-reload';
+
 var socket;
 var watcher;
 
-function cssFileChanged(fileRelativePath) {
-    telemetry.sendTelemetry('live-reload', { fileType: '.css' });
-    socket.emit('lr-update-css', { href: fileRelativePath.replace(/\\/g, '/') });
+function onFileChanged(fileRelativePath, filePathFromProjectRoot) {
```
dlebu: I would break this into multiple functions such that it's a little easier to follow the code.

src/server/live-reload/watcher.js
```diff
@@ -3,19 +3,16 @@
...
 
 var eventIgnoreDuration = 150;
 var wwwRoot = 'www';
 var mergesRoot = 'merges';
 var unknownFileId = '__sim-unknown__';
 
-function Watcher(cssCb, nonCssCb) {
-    this.cssChangedCallback = cssCb;
-    this.nonCssChangedCallback = nonCssCb;
+function Watcher(fileChangedCb) {
```
dlebu: I see you use synchronous file functions in the watcher. Given that the server is used for simulation as well and not just for live reload, would it make sense to switch to async to unblock the server while doing file operations?

dlebu:
I think this looks good. Thanks for making this change. I had a comment about using async file functions in the watcher instead of sync, which shows up as outdated for some reason.

README.md
```diff
@@ -24,6 +24,8 @@ Where opts is an object with the following properties (all optional):
 * **port** - the desired port for the server to use. Defaults to `8000`.
 * **dir** - the directory to launch from (where it should look for a Cordova project). Defaults to cwd.
 * **simhostui** - the directory containing the UI specific files of the simulation host. Defaults to the bundled simulation host files, found in `src/sim-host/ui`.
+* **nolivereload** - A boolean. Set to true to disable live reload. Defaults to false.
```
albertinad: Instead of nolivereload, we can just name this property livereload, and the default value is true.

README.md
```diff
@@ -24,6 +24,8 @@ Where opts is an object with the following properties (all optional):
...
+* **forceprepare** - A boolean. Set to true to force a `cordova prepare` whenever a file changes during live reload. If this is false, the server attempts to simply copy the changed file to the platform rather than doing a `cordova prepare`. Ignored if live reload is disabled. Defaults to false.
```
albertinad: If this configuration is only for live reload functionality, maybe we can define it in a single configuration object in where enabling live reload and forcing prepare can be defined:
```
livereload: {
    enabled: true,
    forceprepare: false
};
```

src/app-host/live-reload-client.js
```diff
@@ -0,0 +1,181 @@
...
+    function findDomNodesForFilePath(fileRelativePath) {
+        // To use querySelectorAll to query elements based on their attributes, the selector's syntax is: '[attrib1], [attrib2], ...'.
+        var selectorString = `[${referenceAttributes.join('], [')}]`;
+        var rawNodes = document.querySelectorAll(selectorString);
+        var filteredNodes = [];
+
+        // querySelectorAll() does not return an array, so we can't use Array.prototype.filter().
+        for (var i = 0; i < rawNodes.length; ++i) {
+            // Ignore <script> tags (we need to do a full reload for scripts).
+            if (rawNodes[i].tagName.toLowerCase() === "script") {
```
albertinad: Just a minor comment: we have mixed double quotes and single quotes. The convention seems to be single quotes.

src/app-host/live-reload-client.js
```diff
@@ -0,0 +1,181 @@
...
+                continue;
+            }
+
+            // Verify if the node is referencing the modified file
+            var referenceAttribute = getReferenceAttributeForNode(rawNodes[i]);
+            var nodeReference = rawNodes[i].getAttribute(referenceAttribute);
```
albertinad: You're using multiple times rawNodes[i]. Maybe you can just save it to a variable, and re-use it.

src/app-host/live-reload-client.js
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Based in part on code from Vogue (https://github.com/andrewdavey/vogue)
+
+var url = require('url');
+
+var canRefreshEventName = 'lr-can-refresh';
+var refreshFileEventName = 'lr-refresh-file';
+var fullReloadEventName = 'lr-full-reload';
```
albertinad: You can define all of this constants in an enumeration
{code}
var EventsName = {
    CAN_REFRESH: 'lr-can-refresh',
    REFRESH_FILE: 'lr-refresh-file',
    FULL_RELOAD: 'lr-full-reload'
};
{code}

src/plugins/cordova-plugin-battery-status/sim-host.js
```diff
@@ -1,14 +1,21 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 
+var telemetry = require('../../client-common/client-telemetry-helper');
```
albertinad: Instead of requiring the module using relative paths, you can define the client-telemetry-helper module to be available in the browserify search path,
```
var telemetry = require('telemetry-helper');
```
So, if this changes the location, you do not need to update all of the plugins simulation files, and you can provide a better infrastructure for developers that wants to develop simulation for their plugins and use telemetry too.

To do this, go to `src/server/sim-files.js`, at the function `getBrowserifySearchPaths`, add the client-common folder to the browserify search path for sim-host,

```
_browserifySearchPaths[simHost] = [dirs.modules[simHost], dirs.modules['common'], dirs.thirdParty]; // here you should add it
```

src/client-common/client-telemetry-helper.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
albertinad: Why this module is defined under client-common? If it is shared between clients like sim-host and/or app-host, then this can be defined under src/modules.

README.md
guillaumejenkins: Done, will be in my next commit

src/client-common/client-telemetry-helper.js
albertinad: Maybe this module can be renamed to telemetry-helper.js instead of the current name.
No matter if this module is going to be under the current folder, or under src/modules, I think it is clear that this is for the client.

README.md
guillaumejenkins: I agree it makes sense for programmatic consumption, but from a command-line perspective, it makes it a bit more awkward to configure live reload. I would be inclined to leave them as separate properties. What does everyone think?

src/app-host/live-reload-client.js
guillaumejenkins: You are right, this is an oversight; it will be fixed in my next commit

src/app-host/live-reload-client.js
guillaumejenkins: Sure, I'll do that, it will be in my next commit

src/app-host/live-reload-client.js
guillaumejenkins: Good idea, it will be in my next commit

src/plugins/cordova-plugin-battery-status/sim-host.js
guillaumejenkins: Great, I didn't know this; it's a much better approach. Will be in my next commit

src/client-common/client-telemetry-helper.js
guillaumejenkins: Great feedback, I wasn't aware module was a common place for clients, I've renamed the telemetry-helper and moved it to modules (will be in my next commit)

src/plugins/cordova-plugin-device-motion/sim-host.js
albertinad: That's a great idea @dlebu! yes, the mobile spec app has tests for all of the core plugins.
You should clone the [mobile spec app repo](https://github.com/apache/cordova-mobile-spec) in some location in which you can clone the other repos needed too. This is the main [documentation](https://github.com/apache/cordova-mobile-spec/blob/master/createmobilespec/README.md) for creating the mobile spec app.

src/plugins/cordova-plugin-geolocation/sim-host.js
```diff
@@ -77,6 +88,66 @@ module.exports = function(messages) {
...
+
+        // Register the event for zooming with the mouse wheel on the map.
+        document.getElementById('geo-map-container').onwheel = mapEventTelemetryHandler;
+    }
+
+    function registerTelemetryForControl(controlId, event) {
+        event = event || 'change';
+        document.getElementById(controlId).addEventListener(event, telemetry.sendUITelemetry.bind(this, Object.assign({}, baseProps, { control: controlId })));
+    }
+
```
albertinad: I see that adding telemetry imply a lot of changes in the plugin simulation code. There are lot of logic related to the telemetry. Maybe, for maintainability and separation of concerns, we can define most of this logic related to telemetry, variables/constants, and helpers functions in other module that expose some APIs to register telemetry or something like that.

src/plugins/cordova-plugin-network-information/sim-host.js
```diff
@@ -30,6 +37,18 @@ function initialize() {
...
+}
+
+function registerTelemetryEvents() {
+    var connectionList = document.getElementById('connection-list');
+
+    connectionList.onchange = function () {
+        var option = connectionList.options[connectionList.selectedIndex];
+
+        telemetry.sendUITelemetry(Object.assign({}, baseProps, { control: 'connection-list', value: option.value }));
+    };
```
albertinad: Now we are sending telemetry events for all of the usages of the simulation plugin. But what are we measuring and what would be the action to perform when we notice the usages of a plugin? Does it makes sense to measure all? What is the main goal of measuring?

src/server/live-reload/watcher.js
```diff
@@ -0,0 +1,93 @@
...
+var config = require('../config');
+var fs = require('fs');
+var log = require('../log');
+var path = require('path');
+var Q = require('q');
+
+var eventIgnoreDuration = 150;
+var wwwRoot = 'www';
+var mergesRoot = 'merges';
+var unknownFileId = '__sim-unknown__';
```
albertinad: If these variables are constants, then we should have some convention about it. Maybe using uppercase letters for the names.

README.md
albertinad: Yes, you're right @guillaumejenkins, after making my comment I realize that it will be good for programmatic API but not so good for the CLI. But it was just a minor comment, as it is right now is good!

src/plugins/cordova-plugin-geolocation/sim-host.js
guillaumejenkins: I agree that it would be ideal, but in practice I find it hard to implement. I tried keeping infrastructure-related code in its own files and modules, but for the actual telemetry points / events, it's hard to be decoupled. We need direct access to the DOM elements so that we can hook up event handlers to some of their events, and giving a separate module access to the DOM might be even more complicated to maintain.

I am fully open to design suggestions that might accomplish what you are describing, but for now I don't know how to easily improve the current approach (having the UI code register its own telemetry events).

src/server/live-reload/watcher.js
guillaumejenkins: You are totally right, I will go over all my changes and capitalize all my constants

src/plugins/cordova-plugin-network-information/sim-host.js
guillaumejenkins: Of course, the questions are totally valid :) Here's what we think:
We believe this will help us understand how the UI is being used, what are the popular plugins, how frequently do people change the simulation values, etc. It will help us make decisions about future investments, how we prioritize plugin simulation work, etc

src/plugins/cordova-plugin-device-motion/sim-host.js
guillaumejenkins: @albertinad Once I create the mobile spec app, all I need to do is launch it in cordova-simulate for the browser platform and watch the console for test failures, correct?

src/plugins/cordova-plugin-network-information/sim-host.js
albertinad: Thanks for your answer @guillaumejenkins! I wonder if it makes sense to keep track of the non-core plugins that doesn't support simulation in order to invest in adding it. Not for all of them but the most popular..

src/plugins/cordova-plugin-device-motion/sim-host.js
albertinad: @guillaumejenkins that's right, but you can launch it for any platform you want to support. I always launch it for android, ios, wp8 and windows. Watching the console will not help so much since there's a lot of information being logged. You can wait for the test report.

src/plugins/cordova-plugin-network-information/sim-host.js
guillaumejenkins: There is also a telemetry event to indicate which plugins are being run with the app. The event reports 3 different list of plugins that the simulated project is using:

- `simulatedBuiltIn`: the list of default cordova-simulate plugins
-`simulatedNonBuiltIn`: the list of non-default plugins for which there was an implementation of the simulation layer
-`notSimulated`: the list of plugins that don't have an implemented simulation

With these 3 lists, we should be able to answer your question about adding new plugins!

albertinad: Thank you! :)

albertinad:
@guillaumejenkins it looks good to me! I just have one minor comment, can you make sure that each file that has been modified has a new line at the end?

PR #14, opened 2016-05-09
Updates to device and media plugins simulation support

albertinad:
Hi! As part of this PR I've added some missing properties to the device plugin simulation:
- isVirtual, default to true for all the devices
- cordova version, which is taken from a method registered in app-host side and called from sim-host side
- serial
- manufacturer

I've also added some missing functions to the media plugin simulation, related to recording audio, those functions are not fully supported yet.

Before this changes,

device-media-before

After this changes,

device-media-after

Thanks!


src/plugins/cordova-plugin-device/sim-host.js
```diff
@@ -312,38 +388,66 @@ function handleSelectDevice() {
...
         registerTelemetryForControl(controlId);
     });
 
-    // Register the event for the tdevice combo box.
+    // Register the event for the device combo box.
     var deviceList = document.getElementById('device-list');
 
-    deviceList.onchange = function () {
+    deviceList.addEventListener('change', function () {
         var option = deviceList.options[deviceList.selectedIndex];
```
albertinad: @guillaumejenkins @dlebu This new event listener for the change event added as part of the telemetry feature, overwrites the previous event listener added using deviceList.onchange, to update the fields according to the device selected. I noticed that when I've updated my branch,  so I fixed here by replacing the usages of onchange with addEventListener, here for the telemetry and for when the device is selected.

dlebu: Sounds good!

dlebu:
This looks good to me, @albertinad!

src/plugins/cordova-plugin-device/sim-host.js
guillaumejenkins: Great catch!

albertinad:
Great! merging!

albertinad:
This close #19

PR #15, opened 2016-05-18
Added XHR proxy support

dlebu:
Summary of the change (based on Apache Ripple's proxy mechanism):

- Added a new `/xhr_proxy` route for the server which handles requests to outside domains
- Modified the implementation of app-host `XMLHttpRequest` to forward requests to `localhost:port/xhr_proxy` when needed
- Added a new `xhrproxy` parameter for simulate which defaults to true


dlebu:
Test slack github integration!

README.md
```diff
@@ -26,6 +26,7 @@ Where opts is an object with the following properties (all optional):
 * **simhostui** - the directory containing the UI specific files of the simulation host. Defaults to the bundled simulation host files, found in `src/sim-host/ui`.
 * **livereload** - A boolean. Set to false to disable live reload. Defaults to true.
 * **forceprepare** - A boolean. Set to true to force a `cordova prepare` whenever a file changes during live reload. If this is false, the server attempts to simply copy the changed file to the platform rather than doing a `cordova prepare`. Ignored if live reload is disabled. Defaults to false.
+* **xhrproxy** - Boolean indicating if a XHR should go through the server proxy. Defaults to true.
```
albertinad: This is for avoiding CORS. I guess that it would be more clear if instead naming the option `xhrproxy`, we use `corsproxy`.

albertinad:
@dlebu Looks good to me. I just have one minor comment.

src/app-host/app-host.js
```diff
@@ -83,6 +84,10 @@ socket.on('init-telemetry', function (data) {
     telemetry.init(socket);
 });
 
+socket.on('init-xhr-proxy', function (data) {
+    xhrProxy.init();
```
albertinad: xhrProxy is needed only when app-host received this message. So maybe, we can load the module, at this point instead of at the top of the module.

src/server/server.js
```diff
@@ -124,6 +126,28 @@ function processPluginHtml(html, pluginId) {
...
+        headers: request.headers
+    };
+    
+    var proxyCallback = function(proxyReponse) {
+        proxyReponse.pipe(response);
+    };
+    
+    http.request(options, proxyCallback).end();
+}
+
```
albertinad: This features, can be done as middlewares. We keep the server as minimum as possible, as a core, and everything we can add as an extra/optional, can be a middleware.
We can implement proxyXHR as part of another module (express middleware), and then use it only if the feature is enabled. As it is right now, we are assuming in the server-side that we need to resolve CORS, when it is not necessary true.

dlebu: I agree. I'll change the implementation to load this lazily, only if the feature is enabled.

README.md
dlebu: Good idea.

dlebu:
@albertinad I addressed all the comments, thanks for the suggestions. Both server and client component are now required as on demand modules. The server proxy component is attached as middleware.

src/simulate.js
```diff
@@ -25,16 +25,23 @@ var launchServer = function (opts) {
...
+    /* attach simulation host middleware */
+    var middlewarePath = path.join(simHostOpts.simHostRoot, 'server', 'server');
+    if (fs.existsSync(middlewarePath + '.js')) {
+        require(middlewarePath).attach(server.app, dirs);
+    }
+    
+    /* attach CORS proxy middleware */
+    if (!!opts.corsproxy) {
+        require('./server/xhr-proxy').attach(server.app);
+    }
```
albertinad: Great!!

albertinad:
Thank you @dlebu! it looks great!

PR #17, opened 2016-05-18
Adding accounting.js to ThirdPartyNotices.txt.

albertinad:
This PR si for updating the third-party notice to include accounting.js library information. This library was added to improve the support of the globalization plugin simulation.


dlebu:
LGTM!

PR #20, opened 2016-05-20
Adding touch events support in app-host

albertinad:
In this PR I've added the support of touch events in app-host. This follows the approach of other configurable functionality, it can be disabled. The implementation is based on Ripple implementation of touch events support.

I've also do a minor refactoring to the config module. Now we have more properties than before, and only one of them has a different setter, so I've changed the way we add properties to the module.export object that is done automatically for most of the properties. I'm open to your feedback about it :)

There's an inconsistency between the documentation about `corsproxy` default value and what it actually does, so I've fixed it to the default value specified in the README.md. @dlebu I can change it back and update the documentation if you think that the default value should be different.

Thanks!

### DIFF
--- a/src/app-host/touch-events.js
+++ b/src/app-host/touch-events.js
@@ -56,11 +56,13 @@ function _simulateTouchEvent(type, mouseevent) {
     eventData.changedTouches.item = itemFn;
     eventData.targetTouches.item = itemFn;
 
-    var simulatedEvent = _createTouchEvent(type, true, true, eventData);
+    var listenerName = 'on' + type,
+        simulatedEvent = _createTouchEvent(type, true, true, eventData);
+
     mouseevent.target.dispatchEvent(simulatedEvent);
 
-    if (typeof mouseevent.target['on' + type] === 'function') {
-        mouseevent.target['on' + type].apply(mouseevent.target, [simulatedEvent]);
+    if (typeof mouseevent.target[listenerName] === 'function') {
+        mouseevent.target[listenerName].apply(mouseevent.target, [simulatedEvent]);
     }
 }
 
--- a/src/app-host/touch-events.js
+++ b/src/app-host/touch-events.js
@@ -1,0 +1,89 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Based in part on code from Apache Ripple, https://github.com/apache/incubator-ripple
+
+var utils = require('utils');
+
+var _lastMouseEvent,
+    _isMouseDown;
+
+// NOTE: missing view, detail, touches, targetTouches, scale and rotation
+function _createTouchEvent(type, canBubble, cancelable, eventData) {
+    var touchEvent = window.document.createEvent('Event');
+    touchEvent.initEvent(type, canBubble, cancelable);
+    utils.mixin(eventData, touchEvent);
+
+    return touchEvent;
+}
+
+function _simulateTouchEvent(type, mouseevent) {
+    if (_lastMouseEvent &&
+            mouseevent.type === _lastMouseEvent.type &&
+            mouseevent.pageX === _lastMouseEvent.pageX &&
+            mouseevent.pageY === _lastMouseEvent.pageY) {
+        return;
+    }
+
+    _lastMouseEvent = mouseevent;
+
+    var touchObj = {
+        clientX: mouseevent.pageX,
+        clientY: mouseevent.pageY,
+        pageX: mouseevent.pageX,
+        pageY: mouseevent.pageY,
+        screenX: mouseevent.pageX,
+        screenY: mouseevent.pageY,
+        target: mouseevent.target,
+        identifier: ''
+    };
+
+    var eventData = {
+        altKey: mouseevent.altKey,
+        ctrlKey: mouseevent.ctrlKey,
+        shiftKey: mouseevent.shiftKey,
+        metaKey: mouseevent.metaKey,
+        changedTouches: [touchObj],
+        targetTouches: type === 'touchend' ? [] : [touchObj],
+        touches: type === 'touchend' ? [] : [touchObj]
+    };
+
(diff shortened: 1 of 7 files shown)


src/modules/utils.js
```diff
@@ -258,4 +266,3 @@ function createUUIDPart(length) {
     }
```
albertinad: What do you think if in the near future we provide a built-in library for this? like loadash or underscore, and we keep our `utils` module only for kind of utility that are not available in those libraries.

src/app-host/touch-events.js
```diff
@@ -0,0 +1,89 @@
...
+    };
+
+    eventData.touches.item = itemFn;
+    eventData.changedTouches.item = itemFn;
+    eventData.targetTouches.item = itemFn;
+
+    var simulatedEvent = _createTouchEvent(type, true, true, eventData);
+    mouseevent.target.dispatchEvent(simulatedEvent);
+
+    if (typeof mouseevent.target['on' + type] === 'function') {
```
dlebu: Nit: I would extract `'on' + type` to a var.

src/app-host/touch-events.js
```diff
@@ -0,0 +1,89 @@
...
+    }, true);
+
+    window.document.addEventListener('mousemove', function (event) {
+        if (_isMouseDown) {
+            _simulateTouchEvent('touchmove', event);
+        }
+    }, true);
+
+    window.document.addEventListener('mouseup', function (event) {
+        _isMouseDown = false;
```
dlebu: I wonder if it's possible to get the `mouseup` event without getting `mousedown` first ( if the user clicks outside of the app host or some other weird scenario like that). If that is possible, I would add the `if (_isMouseDown)` check here as well.

src/modules/utils.js
dlebu: We currently bundle all our dependencies for VSCode and would like to keep our dependency list to a minimum. Whenever we take a new dependency we need to take it to our legal department to verify the license. However, if we find more cases where we need a new library, we are definitely open to it.

src/server/config.js
```diff
@@ -6,71 +6,30 @@ var fs = require('fs'),
...
+    'platform',
+    'platformRoot',
+    'forcePrepare',
+    'projectRoot',
+    'server',
+    'telemetry',
+    'simHostOptions',
+    'xhrProxy',
+    'touchEvents'
+].forEach(function (prop) {
```
dlebu: This is a nice approach, but we are losing our [optional](https://github.com/Microsoft/cordova-simulate/pull/20/files#diff-852bb59fffeac067fe0e2cbf7646bc73R54) flag here. One solution is to remove the [throw call](https://github.com/Microsoft/cordova-simulate/pull/20/files#diff-852bb59fffeac067fe0e2cbf7646bc73L105) as well to avoid errors when querying for an optional config that was not set.

src/simulate.js
```diff
@@ -30,14 +30,15 @@ var launchServer = function (opts) {
     config.telemetry = opts.telemetry;
     config.liveReload = opts.hasOwnProperty('livereload') ? !!opts.livereload : true;
     config.forcePrepare = !!opts.forceprepare;
-    config.xhrProxy = !!opts.corsproxy;
+    config.xhrProxy = opts.hasOwnProperty('corsproxy') ? !!opts.corsproxy : true;
```
dlebu: Thanks for fixing this!

albertinad: No problem :)

src/server/config.js
albertinad: Uh! you're right, I missed that! what do you think about keeping the optional value, and change it in this way:
```
[
    { name: 'liveReload'},
    { name: 'platform' },
    { name: 'platformRoot' },
    { name: 'forcePrepare' },
    { name: 'projectRoot', single: true },
    { name: 'server', single: true },
    { name: 'telemetry', single: true },
    { name: 'xhrProxy', single: true },
    { name: 'simHostOptions' },
    { name: 'touchEvents' }
].forEach(function (prop) {
    Object.defineProperty(module.exports, prop.name, {
        get: function () {
            return getValue(prop.name);
        },
        set: function (value) {
            setValue(prop.name, value, prop.single);
        }
    });
});
```

src/app-host/touch-events.js
albertinad: I will do it, thanks

src/modules/utils.js
albertinad: That's right. Thank you for the clarification!

src/app-host/touch-events.js
albertinad: `mouseup` will only be triggered after the `mousedown` event, that must be triggered over the same element. I did some tests, and wasn't able to get the event. So I guess that we are safe without adding the `_isMouseDown` validation.

src/server/config.js
dlebu: This is a good approach, but we need to add the [optional getter flag](https://github.com/Microsoft/cordova-simulate/pull/20/files/956f9d12da6cede37090e57d5a0b4f473b159213#diff-852bb59fffeac067fe0e2cbf7646bc73L103) as well in addition to the `single` setter flag.

albertinad: @dlebu Today is definitely not my day, I will add the optional flag too and push the changes. Thanks!

src/app-host/touch-events.js
dlebu: Sounds good.

src/server/config.js
dlebu: Sounds good!

albertinad:
@dlebu @guillaumejenkins is this ok to merge? thanks!

dlebu:
:shipit: !

dlebu:
LGTM!

albertinad:
Merged!

PR #21, opened 2016-05-20
Bootstrap protocol implementation

busykai:
Implemented parts:

1. Initial connection;
2. Application reload;
3. Only one `APP_HOST` per instance;
4. Only one `SIM_HOST` per instance;
5. `APP_HOST` reloads on `SIM_HOST` reload.


src/app-host/app-host.js
```diff
@@ -57,38 +100,6 @@ function getCordova() {
     return cordova;
 }
```
albertinad: Plugins simulation (app-host side) is loaded and initialized without being attached to the overall initialization. That could cause issues like requesting exec calls to sim-host side when sim-host is `ready` but plugins are not ready yet because they are initializing. So this exec calls in sim-host fails, because there is no handler, when actually it is.

src/app-host/app-host.js
```diff
@@ -5,41 +5,85 @@ var Messages = require('messages');
...
+
+        socket.on('init-telemetry', function (data) {
+            telemetry.init(socket);
+        });
+
+        socket.on('init-xhr-proxy', function (data) {
+            require('xhr-proxy').init(); 
+        });
+
+        channel.onNativeReady.fire();
```
albertinad: Since plugin's simulation code acts as the native side of the plugin, those should be initialized and ready before firing `onNativeRead`, what do you think @busykai ?

busykai: Yes, `init` message is sent to `APP_HOST` only after `SIM_HOST` tells the server it's ready.

src/app-host/app-host.js
busykai: Thanks for catching this! It was behaving weirdly on me, but I didn't get to actually root-cause it.

src/server/socket.js
```diff
@@ -1,103 +1,204 @@
...
+        .then(onAppHostConnected);
+}
+
+function handleAppPluginList(data) {
+    var socket = hostSockets.SIM_HOST;
 
-            // Set up telemetry if necessary.
-            if (config.telemetry) {
-                socket.emit('init-telemetry');
+    socket.once('start', handleStart);
```
albertinad: This function is called even when there's no sim-host instance connected. Maybe this scenario is not taken into account yet. To reproduce it, launch the simulation, you will get sim-host and app-host connected, then close sim-host so it gets disconnected, and close app-host. Load app-host one more time, and you will si an exception in the server log: `TypeError: Cannot read property 'once' of undefined`.

src/server/socket.js
```diff
@@ -1,103 +1,204 @@
...
+            });
+            if (!type) {
+                log.log('Disconnect for an inactive socket');
+                return;
+            }
+            hostSockets[type] = undefined;
+            log.log(type + ' disconnected to the server');
+            switch (type) {
+                case APP_HOST:
+                    resetAppHostState();
```
albertinad: When app-host is disconnected we reset the state, that means creating a new promise and attach a callback for when the promise is resolved, but when the app-host sends the `register-app-host` message, the promise is resolved. And the handler gets executed assuming that there's a connected sim-host instance.

src/server/socket.js
```diff
@@ -1,103 +1,204 @@
...
+/*
+ * This module implements communication protocol as described
+ */
+
 var log = require('./log'),
     config = require('./config'),
     livereload = require('./live-reload/live-reload-server'),
-    telemetry = require('./telemetry-helper');
+    telemetry = require('./telemetry-helper'),
+    q = require('q');
```
albertinad: The variable name should be with uppercase, for consistency with other usages and since `Q` is the global when no using amd loader.

src/server/socket.js
busykai: thank you for this catch!

src/app-host/app-host.js
```diff
@@ -5,93 +5,184 @@ var Messages = require('messages');
...
+            case 'ubuntu':
+            case 'webos':
+                break;
+            // windows has an overriden bootstrap which does not fire
+            // onNativeReady
+            case 'windows':
+            default:
+                channel.onNativeReady.fire();
+                break;
+        }
```
albertinad: Nice!!

src/server/socket.js
busykai: @albertinad, is this comment related to the crash on when `SIM_HOST` gets connected and then disconnected? I cannot figure out if this is something I addressed already or it's still pending. :)

albertinad: @busykai yes! that is related to that issue. I will try your new changes ASAP :)

src/app-host/app-host.js
```diff
@@ -5,93 +5,184 @@ var Messages = require('messages');
...
-    telemetry.init(socket);
-});
+        socket.on('init-touch-events', function (data) {
+            require('./touch-events').init();
+        });
 
-socket.on('init-xhr-proxy', function (data) {
-    require('xhr-proxy').init();
-});
+        channel.onNativeReady.fire();
```
albertinad: This is duplicated. I guess that you want to fire `onNativeReady` as you're doing below.

busykai: Thanks! That must be rebase artifact.

albertinad:
@busykai, When app-host is reload and about to be sent, sometimes sim-host is requested to be reloaded too. So a `refresh` message is sent to sim-host, and immediately a [function to invalidate sim-host](https://github.com/Microsoft/cordova-simulate/blob/master/src/server/sim-files.js#L43) instance is made. Calling `simSocket.invalidateSimHost();` sets to undefined to the sim-host socket reference that we keep. I think that this is not needed. When sending `refresh`, sim-socket will reload itself, so we are going to receive a disconnection and then, a new sim-host connection.
That call, sometimes causes an error, as you can see in this section of the log:
```
SIM:  Creating app-host.js
SIM:  Emitting 'refresh' to sim-host (pending connection)
Uncaught top-level Node.js exception: TypeError: Cannot read property 'push' of undefined
TypeError: Cannot read property 'push' of undefined
    at Object.emitToHost (<path>/cordova-simulate/src/server/socket.js:268:27)
    at Object.validateSimHostPlugins (<path>/cordova-simulate/src/server/sim-files.js:42:19)
    at <path>/cordova-simulate/src/server/server.js:52:18
From previous event:
    at streamAppHostHtml (<path>/cordova-simulate/src/server/server.js:50:8)
    at Layer.handle [as handle_request] (<path>/components/server/emulator/cordova-simulate/xdk-cordova-serve/node_modules/express/lib/router/layer.js:95:5)
```

busykai:
1. I'll refactor the server code to avoid crashes and print an error instead. It is critical for Intel XDK usecase that server is stable.
2. We no longer need pending emits during the bootstrap as the protocol ensures there must be none.
3. I'll also implement what's left of the protocol spec (except for multiple app-hosts).

albertinad:
That sounds good! thank you @busykai :)

src/app-host/app-host.js
```diff
@@ -5,93 +5,187 @@ var Messages = require('messages');
...
+            console.log(data);
+            var execCacheInfo = execCache[data.index];
+            if (execCacheInfo.success) {
+                execCacheInfo.success(data.result);
+            }
+        });
+
+        socket.on('exec-failure', function (data) {
+            console.log('exec-failure:');
+            console.log(data);
```
albertinad: We can change this to
```
console.log('exec-failure:', data);
```

src/app-host/app-host.js
```diff
@@ -5,93 +5,187 @@ var Messages = require('messages');
...
+            var execCacheInfo = execCache[data.index];
+            if (execCacheInfo.fail) {
+                execCacheInfo.fail(data.error);
+            }
```
albertinad: Does it makes sense to check if `execCacheInfo.fail` is a function, before executing it instead of the current validation?
```
if (typeof execCacheInfo.fail === 'function') {
    execCacheInfo.fail(data.error);
}
```
Also, could we have an scenario in which execCache[data.index] is undefined? If exists such scenario, maybe we can add a better validation too.
I remember that a while ago, there were situations in which execCacheInfo was undefined (in sim-host side), but it was fixed in one of the merges by sending the register event at last.

src/server/socket.js
```diff
@@ -1,115 +1,194 @@
...
 
-            socket.on('debug-message', function (data) {
-                emitToHost(DEBUG_HOST, data.message, data.data);
-            });
+function setupSimHostHandlers() {
+    log.log('Setup handlers for SIM_HOST');
+
+    var socket = hostSockets.SIM_HOST;
+
+    socket.on('exec-success', function (data) {
```
albertinad: Maybe you can use `suscribeTo` function that internally does this for you.

busykai: Yes, it was the idea. I will clean this up with the final review.

src/app-host/app-host.js
busykai: I think it's a good idea to check `execCacheInfo` is actually there. Type check seems to be excessive for me: it would be a plugin simulation bug. We need to simplify a task of adding simulation support for a plugin author, it would be helpful there. I'm thinking of whether it needs to be done now or later (as part of the "developing simulation support" feature later, for example).

Anyways, I agree... I think this should be the condition for the happy scenario:
```javascript
if (execCacheInfo && typeof execCacheInfo.fail === 'function') {
```

src/app-host/app-host.js
busykai: I think we should remove it altogether. I just moved the code around. It's not a minor thing, we should be careful to avoid contaminating application console.

What I think we should add is a warning for `exec()` calls that did not return in a reasonable period of time (5 sec?).

albertinad: That is better! I'm not so sure about the warning in the reasonable period of time. How does cordova handles that? Does the call to native side have a maximum "waiting" time?

busykai: I don't think cordova keeps track of that (among the implementation for the platforms I've seen). We could though, but it will not be part of this PR. We should have a "verbose" mode for `APP_HOST` where things such as exec traces and unreturned calls are printed.

albertinad:
@busykai This looks really great! I've been playing a while and it works. If we find some corner cases or some scenarios that we've missed we could fix it later, but so far, I didn't find anything.

src/server/socket.js
```diff
@@ -1,115 +1,193 @@
...
+        telemetry.handleClientTelemetry(data);
+    });
 
-            socket.on('debug-message', function (data) {
-                emitToHost(DEBUG_HOST, data.message, data.data);
-            });
+    // Set up live reload if necessary.
+    if (config.liveReload) {
+        log.log('Starting live reload.');
+        livereload.init(socket);
```
albertinad: This should be updated too, socket is not available.

albertinad:
@busykai I have one more comment about these changes.
There's a feature that enables simulation code, in both hosts, to emit and listen for events, and also to define methods and execute those methods. There are cases in which the sim-host side of a simulation plugin, needs to get some information that is only available on the app-host side. For example, the device simulation, we need to get the current cordova version, and to do that, we `register` a method in app-host side, that is called by the sim-host side to show the cordova version in the widget.
There's a problem right now: we never get the value, because we're a executing the method before it can be registered or when cordova is not attached to the window yet. With your changes I can propose a fix, but it is not a good one: register a method and also emit an event, and in the sim-host side, call that method, and listen for the event. That is a workaround, but we still need a better solution, and common solution that address all plugin's issues related to this, when they register and execute methods.
Since now we have more clear the steps of the initialization, we know when sim-host is ready, and also when the simulation modules are ready on app-host, so we can queue the messages `plugin-method` (and maybe `plugin-message`?) until both sim-host and app-host are ready, and start to emit them.
What do you think?

src/app-host/app-host.js
```diff
@@ -5,93 +5,184 @@ var Messages = require('messages');
...
+            case 'firefoxos':
+            case 'ubuntu':
+            case 'webos':
+                break;
+            // windows has an overriden bootstrap which does not fire
+            // onNativeReady
+            case 'windows':
+            /* falls through */
+            default:
+                channel.onNativeReady.fire();
```
dlebu: Does `android` go to the default case?

busykai: Yes! Do you think we should explicitly name it here? It might be worth it.

src/server/socket.js
```diff
@@ -1,115 +1,193 @@
...
+
+    subscribeTo(SIM_HOST, 'plugin-method', function (data, callback) {
+        emitTo(APP_HOST, 'plugin-method', data, callback);
+    });
+
+    subscribeTo(SIM_HOST, 'debug-message', function (data) {
+        emitTo(DEBUG_HOST, data.message, data.data);
+    });
+
+    subscribeTo(SIM_HOST, 'telemetry', function (data) {
```
dlebu: Would it make sense to move this `subscribeTo` call in the next `if` statement, such that we subscribe only if telemetry is enabled?

src/server/socket.js
```diff
@@ -1,115 +1,193 @@
...
+
+function init(server) {
+    var io = require('socket.io')(server);
+
+    io.on('connection', function (socket) {
+        socket.on('register-app-host', function () {
+            log.log('APP_HOST connected to the server');
+            if (hostSockets[APP_HOST]) {
+                log.log('Overriding previously connected APP_HOST');
+                resetAppHostState();
```
dlebu: For the current implementation, only 1 APP is supported, right?

src/server/socket.js
```diff
@@ -119,23 +197,48 @@ function init(server) {
...
+                    type = t;
+                }
+            });
+            if (!type) {
+                log.log('Disconnect for an inactive socket');
+                return;
+            }
+            hostSockets[type] = undefined;
+            log.log(type + ' disconnected to the server');
+            switch (type) {
```
dlebu: Do we need to handle DEBUG_HOST in a similar fashion here? @guillaumejenkins

src/sim-host/protocol/socket.js
```diff
@@ -11,52 +11,68 @@ module.exports.initialize = function (pluginHandlers, serviceToPluginMap) {
...
-
-        telemetry.sendClientTelemetry('exec', telemetryProps);
-    });
+
     socket.on('refresh', function () {
         document.location.reload(true);
     });
+
+    socket.on('app-plugin-list', function (data) {
+        // TODO: process the list of plugins
```
dlebu: Please open an issue so we don't forget about this `TODO` if it is still needed.

src/server/socket.js
dlebu: I see that telemetry is sent over the socket even if not enabled, so please ignore my comment if it does not make sense.

src/app-host/app-host.js
dlebu: I think that's a good idea.

dlebu:
The change looks good. I had a couple of comments regarding telemetry and the debug host. I know those features came in after you started this PR, thus I am trying to open the conversation on improving those parts of the code as well.

src/sim-host/protocol/socket.js
```diff
@@ -11,52 +11,68 @@ module.exports.initialize = function (pluginHandlers, serviceToPluginMap) {
...
+            if (!handler) {
+                socket.emit('telemetry', {event: 'exec', props: {handled: 'none', service: service, action: action}});
+                handler = pluginHandlers['*']['*'];
+                handler(success, failure, service, action, data.args);
+            } else {
+                socket.emit('telemetry', {event: 'exec', props: {handled: 'sim-host', service: service, action: action}});
+                handler(success, failure, data.args);
+            }
+
+            telemetry.sendClientTelemetry('exec', telemetryProps);
```
guillaumejenkins: The current implementation of the telemetry-helper under should already take care of that. It will never send telemetry unless it has been told to initialize itself. So, as long as we only call `telemetryHelper.init()` when the server asks it, these calls will not be sent to the server needlessly.

dlebu: Allright, the PR is not up to date with the latest changes from master. Sounds good. Thanks for confirming.

src/server/socket.js
busykai: Yes. I believe I implemented the override here because while testing the refresh/reload scenario I found that we couldn't not rely on old APP_HOST `disconnect` getting before the new APP_HOST `register` message. Perhaps, it could be resolved by closing the socket or explicitly unregistering in `onbeforeunload`, but the new-registration-overrides type of policy is the safest.

src/sim-host/protocol/socket.js
```diff
@@ -8,77 +8,93 @@ var registerOnInitialize = false;
...
+                    throw 'Exec called on simulation host without exec info';
+                }
+
+                index = data.index;
+                if (typeof index !== 'number') {
+                    throw 'Exec called on simulation host without an index specified';
+                }
+
+                var success = data.hasSuccess ? getSuccess(index) : null;
+            var failure = data.hasFail ? getFailure(index) : null;
```
albertinad: Fix indentation here.

albertinad:
This looks good to me! I believe that we are really good to merge it to master, and if we find areas of improvements, we can do it from there. We've been using this new protocol for a while, and all major issues have been addressed and it works really great!

dlebu:
LGTM! :shipit:

src/app-host/app-host.js
guillaumejenkins: Since `windows` is already there with a `/* falls through */` comment, I think it's a good idea to add `android` as well

src/server/socket.js
busykai: I moved the `DEBUG_HOST` disconnect handler here.

src/sim-host/protocol/socket.js
busykai: I need to fix linting to include some formatting checks... Half of the file has been indented wrongly.

src/app-host/app-host.js
```diff
@@ -7,10 +7,61 @@ var telemetry = require('telemetry-helper');
...
+                plugins[pluginId] = plugin;
+            }
+            if (clobberScope) {
+                clobber(plugin, clobberScope, clobberToPluginMap, pluginId);
+            }
+        }
+    });
+}
 
 function setCordova(originalCordova) {
```
guillaumejenkins: Should we rename the function now that it does pretty much all the app-host initialization? The name suggests it only overrides the existing cordova, when in reality it does several extra things (socket handlers for exec calls, telemetry, XHR, touch events and live reload, initializing plugins, etc)... Or, maybe refactor all the non-cordova stuff to a new function? That way we would have a clear separation of cordova initialization and app-host-specific initialization... Something like `initCordova()` and `initAppHost()`?

src/server/socket.js
guillaumejenkins: Telemetry should not be sent over the socket if not enabled. The `init-telemetry` message below is needed by the client side for telemetry to start being sent. For that reason, I think it does make sense to move the subscribe inside the `if`

busykai:
@guillaumejenkins, this could seem a bit counter-intuitive, I agree. And it potenitally could be improved (in sense of code readability). However, we inevitably have to "hack" cordova because it does not have mechanisms in place which would a better designed/structured approach to simulation initialization. `app-host.js` is injected at the very beginning of the head and is therefore executed first which makes patching and the rest of it possible (since it is loaded and exoverride the setter for a global).

Parallel initialization cannot be done since the protocol requires a `cordova` instance to be present already (it references it). Note that neither `cordova` patching nor socket handlers setup are a deferred action, so use of promises is not needed. Setting `cordova` is an action in the future for which we could use promise.

So we could refactor to do something like:

```javascript
function init() {
    // setup protocol socket listeners
    socket.emit('register-app-host');
}

function setCordova(cordova) {
    // patch cordova
    // create sim channel and delay deviceReady
   init();
}
```
I was thinking about something like this. However, it will be even more confusing to me. IMO it's incorrect to think that protocol setup and patching `cordova` are two different and independent activities. Also, we need to `init()` to execute synchronously (since it subscribes to some `cordova` events). 

About this specific point:
> I think it also takes care of a race condition: in your proposed changes, what happens if the server emits init to app-host before cordova-serve does `window.cordova = ...`?

Server takes care of all the timing-sensitive messages. `init` will never be sent before `windows.cordova = ...` since `register-app-host` is only send upon assigning an object to `window.cordova`. Event sequencing is quite fine-tuned by now.

busykai:
@guillaumejenkins, you're welcome! i hope they make sense. i've renamed the function to `setCordovaAndInitialize`. are you OK with the name?

I'll test it a little bit more over the weekend then it will be ready to merge.

busykai:
@guillaumejenkins, @dlebu, @albertinad I've addressed all the comments except for timing of `plugin-*` messages (which I'm working on right now). I'll merge this PR by tomorrow.

busykai:
I'm going to merge it as it as and file a separate issue to address plugins' message timing.

dlebu:
Sounds good!

busykai:
Filed #90. (BTW, allow me to take it.) Merging this one.

albertinad:
Awesome!!

PR #23, opened 2016-05-25
CoreAndroid.backHistory app-host implementation and minor fixes.

albertinad:
Hi! This PR is for adding the JS implementation of the native call to `CoreAndroid.backHistory` that is executed when `navigator.app.backHistory()` is called.
I also remove some unused variables in all of the modules and replace double quotes with single quotes. There's no well-defined convention about this, but since in the majority of the modules we are using single quotes, I replace them with single quotes.

Thanks!


dlebu:
This looks good to me! We should probably add a JS linter to our project and define some coding guidelines in the near future.

albertinad:
Yes! @busykai was working on adding ESLint with Gulp, as a task manager, I will check the status of that.

dlebu:
Sounds good. We use gulp a lot in our typescript projects.

PR #25, opened 2016-05-25
Support for debug-host and viewport resizing

guillaumejenkins:
- Added support for an external debug-host to register with the server, specifying which debug messages it can handle
- Added support for plugins to activate based on the presence of a debug-host and specific debug-host message handlers
- Added the viewport plugin, which allows users to request viewport resizing of their running app (requires an external debug-host to do the actual resizing; this plugin merely sends messages to the connected debug-host)

Please note that for now, the viewport plugin won't sync up with the device plugin (changing a predefined device in one will not update the other). This can be tracked as an Enhancement issue and be done at a later time.

### DIFF
--- a/src/plugins/viewport/sim-host-panels.html
+++ b/src/plugins/viewport/sim-host-panels.html
@@ -1,0 +1,12 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
+
+<cordova-panel id="viewport" caption="Viewport (Screen Size)">
+    <cordova-group id="viewport-options">
+        <cordova-radio id="viewport-desktop" checked="true">Use full browser window</cordova-radio>
+        <cordova-radio id="viewport-device">Use predefined screen size:</cordova-radio>
+        <cordova-combo id="viewport-device-list"></cordova-combo>
+        <cordova-radio id="viewport-custom">Use these screen dimensions:</cordova-radio>
+        <cordova-text-entry label="Width:" id="viewport-custom-width"></cordova-text-entry>
+        <cordova-text-entry label="Height:" id="viewport-custom-height"></cordova-text-entry>
+    </cordova-group>
+</cordova-panel>
--- a/src/plugins/viewport/sim-host.js
+++ b/src/plugins/viewport/sim-host.js
@@ -1,0 +1,193 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+
+var telemetry = require('telemetry-helper');
+
+var baseProps = {
+    plugin: 'viewport',
+    panel: 'viewport'
+};
+var devices = [
+    { 'id': 'AcerA500', 'name': 'Acer A500', 'width': 800, 'height': 1280 },
+    { 'id': 'Bold9700', 'name': 'BlackBerry Bold 9700', 'width': 480, 'height': 360 },
+    { 'id': 'Bold9900', 'name': 'BlackBerry Bold 9900', 'width': 640, 'height': 480 },
+    { 'id': 'Curve9300', 'name': 'BlackBerry Curve 9300', 'width': 320, 'height': 240 },
+    { 'id': 'Curve9350-9360-9370', 'name': 'BlackBerry Curve 9350/9360/9370', 'width': 480, 'height': 360 },
+    { 'id': 'FWVGA', 'name': 'Generic - FWVGA (480x854)', 'width': 480, 'height': 854 },
+    { 'id': 'G1', 'name': 'HTC G1', 'width': 320, 'height': 480 },
+    { 'id': 'HPPre3', 'name': 'HP Pre 3', 'width': 480, 'height': 800 },
+    { 'id': 'HPVeer', 'name': 'HP Veer', 'width': 320, 'height': 400 },
+    { 'id': 'HVGA', 'name': 'Generic - HVGA (320x480)', 'width': 320, 'height': 480 },
+    { 'id': 'iPad', 'name': 'iPad', 'width': 768, 'height': 1024 },
+    { 'id': 'iPad3', 'name': 'iPad 3', 'width': 1536, 'height': 2048 },
+    { 'id': 'iPhone3', 'name': 'iPhone 3G/3Gs', 'width': 320, 'height': 480 },
+    { 'id': 'iPhone4', 'name': 'iPhone 4/4s', 'width': 640, 'height': 960 },
+    { 'id': 'iPhone5', 'name': 'iPhone 5', 'width': 640, 'height': 1136 },
+    { 'id': 'Legend', 'name': 'HTC Legend', 'width': 320, 'height': 480 },
+    { 'id': 'Nexus', 'name': 'Nexus One', 'width': 480, 'height': 800 },
+    { 'id': 'Nexus4', 'name': 'Nexus 4', 'width': 768, 'height': 1280 },
+    { 'id': 'Nexus7', 'name': 'Nexus 7 (Tablet)', 'width': 800, 'height': 1280 },
+    { 'id': 'NexusGalaxy', 'name': 'Nexus (Galaxy)', 'width': 720, 'height': 1280 },
+    { 'id': 'NexusS', 'name': 'Nexus S', 'width': 480, 'height': 800 },
+    { 'id': 'NokiaN8', 'name': 'Nokia N8', 'width': 360, 'height': 640 },
+    { 'id': 'NokiaN97', 'name': 'Nokia N97/5800 (touch)', 'width': 360, 'height': 640 },
+    { 'id': 'PalmPre', 'name': 'Palm Pre', 'width': 320, 'height': 480 },
+    { 'id': 'PalmPre2', 'name': 'Palm Pre 2', 'width': 320, 'height': 480 },
+    { 'id': 'Pearl9100', 'name': 'BlackBerry Pearl 9100', 'width': 360, 'height': 400 },
+    { 'id': 'Playbook', 'name': 'BlackBerry Playbook', 'width': 1024, 'height': 600 },
+    { 'id': 'Q10', 'name': 'BlackBerry Q10', 'width': 720, 'height': 720 },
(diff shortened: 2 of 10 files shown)


src/plugins/cordova-plugin-device-orientation/sim-host.js
```diff
@@ -32,7 +32,7 @@ module.exports = function (messages) {
         });
         compassWidget.initialize(compass.heading);
 
-        inputHeading.addEventListener('input', function () {
+        inputHeading.addEventListener('change', function () {
```
guillaumejenkins: Unrelated telemetry change; the `input` event was causing the value to change more than once while typing a 2- or 3-digit number. With the `change` event, typing `256` in the text entry will not trigger 3 separate events (one for `2`, one for `25` and one for `256`), so I think it's a reasonable improvement

src/plugins/cordova-plugin-device/sim-host.js
```diff
@@ -429,12 +429,17 @@ function registerTelemetryEvents() {
...
-    // Register event for the isVirtual device checkbox
-    document.getElementById('is-virtual-device')
-        .addEventListener('click',
-            telemetry.sendUITelemetry.bind(this, Object.assign({}, baseProps, {
-                control: 'is-virtual-device'
-            })));
+    // Register event for the isVirtual device checkbox. Clicking the checkbox's label fires the click event twice, so
+    // keep track of the previous state. Note that we can't use the change event because the component seems to
+    // swallow it.
+    var previousVirtualState = true;
```
guillaumejenkins: Unrelated telemetry change; the `click` event on `cordova-checkbox` components gets fired twice when clicking the checkbox's label. This logic is needed to work around that.

src/plugins/viewport/sim-host.js
```diff
@@ -0,0 +1,173 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+
+var telemetry = require('telemetry-helper');
+
+var baseProps = {
+    plugin: 'viewport',
+    panel: 'viewport'
+};
+var devices = [
+    {
```
guillaumejenkins: Temporary list; I will integrate @albertinad's device list later today

src/plugins/viewport/sim-host-panels.html
```diff
@@ -0,0 +1,12 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
+
+<cordova-panel id="viewport" caption="Viewport (Device Screen)">
+    <cordova-group id='viewport-options'>
```
dlebu: Nit: Please use double quotes for consistency.

src/plugins/cordova-plugin-device/sim-host.js
```diff
@@ -429,12 +429,17 @@ function registerTelemetryEvents() {
...
+    var virtualDeviceCheckbox = document.getElementById('is-virtual-device');
+    virtualDeviceCheckbox.addEventListener('click', function () {
+        if (virtualDeviceCheckbox.checked !== previousVirtualState) {
+            previousVirtualState = virtualDeviceCheckbox.checked;
+            telemetry.sendUITelemetry(Object.assign({}, baseProps, { control: 'is-virtual-device' }));
```
dlebu: Do we still need to `.bind()` here?

src/plugins/viewport/sim-host-panels.html
guillaumejenkins: Woops :)

src/plugins/viewport/sim-host.js
```diff
@@ -0,0 +1,193 @@
...
+            registerTelemetryForControl(controlId);
+        });
+
+        // Register the event for the device combo box.
+        var deviceList = document.getElementById('viewport-device-list');
+
+        deviceList.addEventListener('change', function () {
+            var option = deviceList.options[deviceList.selectedIndex];
+
+            telemetry.sendUITelemetry(Object.assign({}, baseProps, {
```
dlebu: Here as well. Do we need to bind?

src/plugins/cordova-plugin-device/sim-host.js
guillaumejenkins: We don't, as we are invoking the function ourselves

We need to use `bind` when the function is invoked as an event handler, because the `bind` allows us to send extra arguments that the event doesn't send. Here, we can send whatever arguments we want

src/plugins/cordova-plugin-device-orientation/sim-host.js
busykai: It would be highly desirable to add these changes in a separate commits. As you spot these, you could `git add -p` and `git commit` them in situ so they don't mix with the changes of other nature. It would make bisecting later so much easier and the commits more atomic. Then we don't have to go back to PR comments to see why they appeared in the code in a middle of something that does not seem quite relevant.

guillaumejenkins: Right, I'll remove these changes from here and add them to my other active branch where I'm already doing telemetry changes

Thanks!

src/plugins/viewport/sim-host.js
```diff
@@ -0,0 +1,193 @@
...
+    { 'id': 'QVGA', 'name': 'Generic - QVGA (240X320)', 'width': 240, 'height': 320 },
+    { 'id': 'Style9670', 'name': 'BlackBerry Style 9670', 'width': 360, 'height': 400 },
+    { 'id': 'Tattoo', 'name': 'HTC Tattoo', 'width': 240, 'height': 320 },
+    { 'id': 'Torch9800', 'name': 'BlackBerry Torch 9800', 'width': 360, 'height': 480 },
+    { 'id': 'Torch9810', 'name': 'BlackBerry Torch 9810', 'width': 480, 'height': 640 },
+    { 'id': 'Torch9860-9850', 'name': 'BlackBerry Torch 9860/9850', 'width': 480, 'height': 800 },
+    { 'id': 'Wave', 'name': 'Samsung Wave', 'width': 480, 'height': 800 },
+    { 'id': 'WQVGA', 'name': 'Generic - WQVGA (240x480)', 'width': 240, 'height': 400 },
+    { 'id': 'WVGA', 'name': 'Generic - WVGA (480x800)', 'width': 480, 'height': 800 },
+    { 'id': 'Z10', 'name': 'BlackBerry Z10', 'width': 768, 'height': 1280 }
```
albertinad: We could handle this list of devices in a different way. This is metadata, there's no need to be defined here in the code, it can be a separate json or multiple's json files with all the required information. When more information is required, this will grow, even when we add new devices to support. Having them in separate files, or at leas in a single dedicated file will help for maintainability. 
For today this is OK, but in the future we will need a better handling of devices metadata.

src/simulate.js
```diff
@@ -59,7 +59,7 @@ var launchServer = function (opts) {
         appUrl = urlRoot + parseStartPage(projectRoot);
         simHostUrl = urlRoot + 'simulator/index.html';
         log.log('Server started:\n- App running at: ' + appUrl + '\n- Sim host running at: ' + simHostUrl);
-        return { appUrl: appUrl, simHostUrl: simHostUrl };
+        return { urlRoot: urlRoot, appUrl: appUrl, simHostUrl: simHostUrl };
```
albertinad: Just a minor comment, should it be rootUrl instead of urlRoot?

src/plugins/viewport/sim-host.js
guillaumejenkins: Great point; yes, I had thought of that, but my stance on it was that this should happen as part of our efforts to synchronize the Device and the Viewport plugins, since they will need to share device metadata. To make sure we keep that in mind, I have opened an issue to track this (#27). Thanks!

src/simulate.js
guillaumejenkins: I think they both make sense (`urlRoot` = "the root of all URLs served by the server" versus `rootUrl` = "the URL pointing to the root of the server"), but since Tim had already used `urlRoot`, I'm inclined to just leave it like this :) Is that fine?

albertinad: @guillaumejenkins is more than fine :)!

src/server/plugins.js
```diff
@@ -120,6 +161,63 @@ function findBuiltInPluginSourceFilePath(pluginId, file) {
...
 }
 
+function shouldUsePluginWithDebugHost(pluginPath, pluginAddedToProject) {
+    pluginAddedToProject = !!pluginAddedToProject;
+
+    // Check whether the plugin defines some debug-host requirements.
+    var debugHostFilePath = path.join(pluginPath, pluginSimulationFiles['debug-host']['debug-host'])
+
+    if (!fs.existsSync(debugHostFilePath)) {
+        if (path.basename(pluginPath) === "viewport") { console.log("-----Didn't find debug-host.json"); } // TODO REMOVE
```
dlebu: Still need to remove this line?

src/server/plugins.js
```diff
@@ -120,6 +161,63 @@ function findBuiltInPluginSourceFilePath(pluginId, file) {
...
+    }
+
+    var debugHostOptions;
+
+    try {
+        debugHostOptions = JSON.parse(fs.readFileSync(debugHostFilePath, 'utf8'));
+    } catch (err) {
+        // Can't determine plugin's debug-host requirements, so ignore them. Use the plugin if it was added to the
+        // project.
+        if (path.basename(pluginPath) === "viewport") { console.log("-----Couldn't read debug-host.json"); } // TODO REMOVE
```
dlebu: Same here

dlebu:
Minor comments. This looks good to me!

src/server/plugins.js
guillaumejenkins: Oops, good catch!

src/plugins/viewport/sim-host.js
TimBarham: Speaking of which - are you guys going to look at synchronizing Device and platform? That is, if I choose an Android device in the devices panel and simulate was launched for 'windows' platform, it should switch to Android platform (this is what Ripple does). Similarly when launching for a particular platform, it should choose an appropriate device for that platform (or remember the last used device for a platform).

PR #26, opened 2016-05-27
Telemetry improvements and additions

guillaumejenkins:
- Added telemetry for unhandled exec calls
- Minor stabilizations in existing telemetry

### DIFF
--- a/src/modules/telemetry-helper.js
+++ b/src/modules/telemetry-helper.js
@@ -1,23 +1,63 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 
 var clientSocket;
+var serviceToPluginMap;
+var pendingTelemetryEvents = [];
 
 function init(socket) {
     clientSocket = socket;
+    trySendPendingEvents();
 }
 
-function sendClientTelemetry(event, props, piiProps) {
+function registerPluginServices(pluginServices) {
+    serviceToPluginMap = pluginServices;
+    trySendPendingEvents();
+}
+
+function trySendPendingEvents() {
     if (!clientSocket) {
         return;
     }
 
-    var telemetryData = {
+    var unsent = [];
+
+    pendingTelemetryEvents.forEach(function (eventData) {
+        if (mustMapServiceToPlugin(eventData) && !serviceToPluginMap) {
+            unsent.push(eventData);
+        } else {
+            sendClientTelemetry(eventData);
+        }
+    });
+    
+    pendingTelemetryEvents = unsent;
+}
+
+function mustMapServiceToPlugin(eventData) {
+    return !!eventData.props.service && !eventData.plugin;
+}
+
+function sendClientTelemetry(event, props, piiProps) {
+    var eventData = {
         event: event,
         props: props,
         piiProps: piiProps
     };
 
-    clientSocket.emit('telemetry', telemetryData);
+    if (!clientSocket) {
+        pendingTelemetryEvents.push(eventData);
+        return;
+    }
+
+    if (mustMapServiceToPlugin(eventData)) {
+        if (!serviceToPluginMap) {
+            pendingTelemetryEvents.push(eventData);
+            return;
+        }
+
+        eventData.props.plugin = serviceToPluginMap[eventData.props.service] || '_unknown';
+    }
+
+    clientSocket.emit('telemetry', eventData);
 }
 
 function sendUITelemetry(uiControlData) {
@@ -25,5 +65,6 @@ function sendUITelemetry(uiControlData) {
 }
 
 module.exports.init = init;
+module.exports.registerPluginServices = registerPluginServices;
 module.exports.sendClientTelemetry = sendClientTelemetry;
 module.exports.sendUITelemetry = sendUITelemetry;
--- a/src/plugins/exec/sim-host-handlers.js
+++ b/src/plugins/exec/sim-host-handlers.js
@@ -1,7 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 
-var savedSims = require('./saved-sims');
 var dialog = require('dialog');
+var savedSims = require('./saved-sims');
+var telemetry = require('telemetry-helper');
 
 // Handle any calls not handled by anything else...
 module.exports = {
@@ -24,7 +25,7 @@ module.exports = {
                     // Display of the dialog was delayed for some reason. Check if in the meantime we have a saved value
                     // for this call.
                     return !handleSavedSim(success, fail, service, action);
-                } else  if (msg === 'showing') {
+                } else if (msg === 'showing') {
                     // Prepare the dialog for showing
                     var successButton = document.getElementById('exec-success');
                     var failureButton = document.getElementById('exec-failure');
@@ -45,6 +46,7 @@ module.exports = {
 
                     function exec(func, isSuccess) {
                         var result = resultField.value;
+                        var shouldPersist = document.getElementById('exec-persist').checked;
 
                         try {
                             result = result && JSON.parse(result);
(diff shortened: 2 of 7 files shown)


src/modules/telemetry-helper.js
```diff
@@ -1,29 +1,70 @@
...
+        return;
+    }
+
+    if (mustMapServiceToPlugin(eventData)) {
+        if (!serviceToPluginMap) {
+            pendingTelemetryEvents.push(eventData);
+            return;
+        }
+
+        eventData.props.plugin = serviceToPluginMap[eventData.props.service] ? serviceToPluginMap[eventData.props.service] : '_unknown';
```
albertinad: This could be:
```
eventData.props.plugin = serviceToPluginMap[eventData.props.service] || '_unknown';
```

guillaumejenkins: Indeed! Done

src/modules/telemetry-helper.js
```diff
@@ -1,29 +1,70 @@
...
+
+function mustMapServiceToPlugin(eventData) {
+    return !!eventData.props.service && !eventData.plugin;
+}
+
+function sendClientTelemetry(event, props, piiProps) {
+    var eventData = {
         event: event,
         props: props,
         piiProps: piiProps
```
albertinad: What does `piiProps` means?

src/plugins/exec/sim-host-handlers.js
```diff
@@ -87,3 +96,19 @@ function handleSavedSim(success, fail, service, action) {
...
+    var props = {
+        'service': service,
+        'action': action,
+        'has-persisted-result': hasPersisted,
+        'is-success': isSuccess,
+        'has-result-value': hasResult
+    };
+
+    if (typeof saveResult !== 'undefined') {
+        props['save-result'] = saveResult;
```
albertinad: What is the expected type of `saveResult`? From what I saw it seems to be a boolean, so maybe you can do the following:
```
props['save-result'] = !!saveResult; // defaults to false
```
Why this property is not always available in the `props` object? Is there any specific reason to not to do it?

src/plugins/exec/sim-host-handlers.js
```diff
@@ -87,3 +96,19 @@ function handleSavedSim(success, fail, service, action) {
     }
     return false;
 }
+
+function sendExecUnhandledTelemetry(service, action, hasPersisted, isSuccess, hasResult, saveResult) {
```
albertinad: It would be really helpful if we add some jsdoc to the functions, at least the parameters types, and which of them are optional if any.

src/modules/telemetry-helper.js
guillaumejenkins: PII is acronym for "Personally Identifiable Information". In case of telemetry, we want to separate what is "sensitive" info and what is not, so we put them in `piiProps` and `props` respectively.

Please note that cordova-simulate does not do anything special with PII data; it's up to the programmatic consumer to decide how to deal with it. In case of our cordova extension, for example, when `piiProps` is not empty, we hash all the values contained in it so that it's impossible to know what the real values are (even for us).

As to why we consider plugins to be PII, it's because Cordova supports custom plugins, which can be installed from local paths, which can expose information about the user and his machine. So, for any plugin that is not built-in, we use `piiProps` to notify the programmatic consumer that it should anonymize the properties it contains.

Hope this helps!

albertinad: Thank you for the clarification @guillaumejenkins! :) It helps a lot. I didn't know about the acronym. Maybe it could be worth to add some documentation, same as you explain to me, to help developers to understand the usages of this, and also for those who wants to contribute to the project.

src/plugins/exec/sim-host-handlers.js
guillaumejenkins: I feel like in this case, the absence of `save-result` is not the same as `save-result = false`. The reason is that if `has-persisted-value` is `true`, then `save-result` should be ignored because the user wasn't even presented the choice of whether to persist the call or not. The dialog box didn't even pop up, because that call result was already persisted.

So, I was just afraid that default this to `false` might be misleading when querying telemetry data. For example, searching on "WHERE save-result = FALSE" might give the impression that 75% of the users chose not to persist their unhandled exec call, when in reality half of those weren't even presented the choice.

I hope I'm not being too confusing :) I'll add JSDoc headers to explain the parameters, which should help!

src/plugins/exec/sim-host-handlers.js
guillaumejenkins: Added headers!

src/plugins/exec/sim-host-handlers.js
```diff
@@ -56,10 +58,13 @@ module.exports = {
 
                         dialog.hideDialog('exec-dialog');
 
-                        if (document.getElementById('exec-persist').checked) {
-                            savedSims.addSim({service: service, action: action, args: args, value: result, success: isSuccess});
+                        if (shouldPersist) {
+                            savedSims.addSim({ service: service, action: action, args: args, value: result, success: isSuccess });
                         }
 
+                        var hasValue = typeof result !== 'undefined' && result !== '';
```
dlebu: Can result ever be `null` here?

src/plugins/exec/sim-host-handlers.js
```diff
@@ -78,7 +83,11 @@ module.exports = {
 function handleSavedSim(success, fail, service, action) {
     var savedSim = savedSims.findSavedSim(service, action);
     if (savedSim) {
-        if (savedSim.success) {
+        var isSuccess = !!savedSim.success;
+        var hasValue = typeof savedSim.value !== 'undefined' && savedSim.value !== '';
```
dlebu: We could create a utility method for this.

dlebu:
A couple of comments, this looks good to me.

src/plugins/exec/sim-host-handlers.js
guillaumejenkins: Yes, and that means that the user typed `null` in the popup dialog. In that case, I think that we should consider `hasValue` to be true, because the user entered a value, and we are interested in how users use the UI.

On the other hand, the current implementation prevents users from typing `undefined` in the popup, so the current check works in that case (if `typeof result === 'undefined'` then it truly means the user didn't type anything).

src/sim-host/protocol/socket.js
```diff
@@ -7,6 +7,10 @@ var socket;
 module.exports.initialize = function (pluginHandlers, serviceToPluginMap) {
     socket = io();
     module.exports.socket = socket;
+    module.exports.notifyPluginsReady = function () {
+        telemetry.registerPluginServices(serviceToPluginMap);
+        socket.emit('register-simulation-host');
+    }
```
albertinad: You can ignore this message :) I guess that it can be easy to read and get a description of the public/exported API of a module, if that is defined at the end, or at module level. This module is not so big, but if it grows, then It could be harder to find what are the functions that this module exports. I've seen in other modules of this project, declarations like the following:
```
Object.defineProperty(module.exports, 'socket', {
    get: function () {
        return socket; // it can be undefined if it is called before initialize
    }
});
module.exports.initialize = function () {
    // ... 
};
module.exports.notifyPluginsReady = function () {
    // same as you define but maybe to avoid calling this function before initailize, validate for socket instance
};
```
Maybe with this type you can achieve the same result.

albertinad:
@guillaumejenkins This looks good to me! I just only have one more minor comment, nothing important :) thanks!

src/sim-host/protocol/socket.js
guillaumejenkins: You are right, this will make the code much more maintainable, so I've made these changes!

PR #29, opened 2016-05-27
Fixing a minor config issue.

dlebu:
Attaching the XHR proxy in the default (enabled) case as well.


albertinad:
Looks good to me!

PR #33, opened 2016-05-31
cordova-plugin-inappbrowser simulation support

albertinad:
Hi! This PR is to add support for the InAppBrowser plugin simulation. It supports all of the available targets for loading a URL using InAppBrowser:
- `_self`: It loads the url in the current window, simulating the navigation to the same WebView
- `_blank`: It creates a div with an iframe and a close control that will show up on top of the main window, taking the full size. This acts as creating a new WebView to navigate to the given URL.
- `_system`: It opens the URL in a new window.
For the case of `_blank` target, all actions are supported, even injecting code into the iframe. But that depends on the URL being loaded on the iframe, for security reasons. So we try to inject JS code and CSS styles as we can, otherwise, an error is logged. This was based on the browser proxy implemented for the plugin.

screen shot 2016-05-31 at 2 40 28 am

screen shot 2016-05-31 at 2 41 06 am

I want to share this with you to get feedback, in the meantime I will continue testing it with the mobilespec app to improve it.

PS: If you want to try it out, this change #31 needs to be added in order to successfully execute app-host side handlers.
 
Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ This preview version currently includes built-in support for the following Cordo
 * [cordova-plugin-file](https://github.com/apache/cordova-plugin-file)
 * [cordova-plugin-geolocation](https://github.com/apache/cordova-plugin-geolocation)
 * [cordova-plugin-globalization](https://github.com/apache/cordova-plugin-globalization)
+* [cordova-plugin-inappbrowser](https://github.com/apache/cordova-plugin-inappbrowser)
 * [cordova-plugin-media](https://github.com/apache/cordova-plugin-media)
 * [cordova-plugin-network-information](https://github.com/apache/cordova-plugin-network-information)
 * [cordova-plugin-vibration](https://github.com/apache/cordova-plugin-vibration)
--- a/src/plugins/cordova-plugin-inappbrowser/app-host-handlers.js
+++ b/src/plugins/cordova-plugin-inappbrowser/app-host-handlers.js
@@ -1,0 +1,54 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+
+module.exports = function () {
+    var IframeBrowser = require('./inapp-browser').IframeBrowser,
+        SystemBrowser = require('./inapp-browser').SystemBrowser,
+        instance;
+
+    return {
+        'InAppBrowser': {
+            'open': function (success, fail, args) {
+                // args[0]: url, args[1]: target, args[2]: options
+                var Constructor;
+
+                switch (args[1]) {
+                case '_system':
+                    // open in a new browser tab
+                    Constructor = SystemBrowser;
+                    break;
+                case '_blank':
+                    Constructor = IframeBrowser;
+                    break;
+                default:
+                    // "_self" and any other option, use the same tab
+                    window.location = args[0];
+                    return;
+                }
+
+                instance = new Constructor(args[0], args[2], success, fail);
+            },
+            'show': function () {
+                if (instance) {
+                    instance.show();
+                }
+            },
+            'close': function () {
+                if (instance) {
+                    instance.close();
+                }
+            },
+            'injectScriptCode': function (success, fail, args) {
+                instance.injectScriptCode(success, args);
+            },
+            'injectScriptFile': function (success, fail, args) {
+                instance.injectScriptFile(success, args);
+            },
+            'injectStyleCode': function (success, fail, args) {
+                instance.injectStyleCode(success, args);
+            },
+            'injectStyleFile': function (success, fail, args) {
+                instance.injectStyleFile(success, args);
+            }
+        }
+    };
+};
--- a/src/plugins/cordova-plugin-inappbrowser/app-host-handlers.js
+++ b/src/plugins/cordova-plugin-inappbrowser/app-host-handlers.js
@@ -1,8 +1,17 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 
-module.exports = function () {
+module.exports = function (messages) {
     var Browser = require('./inapp-browser'),
         instance;
+
+    messages.call('inAppBrowserSelected')
+        .then(function (value) {
+            Browser.setDefaultInAppBrowser(value);
+        });
+
+    messages.on('inappbrowser-selected', function (event, value) {
+        Browser.setDefaultInAppBrowser(value);
+    });
 
     function execute(fnName) {
         if (!instance) {
--- a/src/plugins/cordova-plugin-inappbrowser/app-host-handlers.js
+++ b/src/plugins/cordova-plugin-inappbrowser/app-host-handlers.js
@@ -31,6 +31,7 @@ module.exports = function () {
             },
             'close': function () {
                 execute('close');
+                instance = null;
             },
             'injectScriptCode': function (success, fail, args) {
                 execute('injectScriptCode', success, args);
--- a/src/plugins/cordova-plugin-inappbrowser/app-host-handlers.js
+++ b/src/plugins/cordova-plugin-inappbrowser/app-host-handlers.js
@@ -1,53 +1,48 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 
 module.exports = function () {
-    var IframeBrowser = require('./inapp-browser').IframeBrowser,
-        SystemBrowser = require('./inapp-browser').SystemBrowser,
+    var Browser = require('./inapp-browser'),
         instance;
+
+    function execute(fnName) {
+        if (!instance) {
+            return;
+        }
+
+        var fn = instance[fnName],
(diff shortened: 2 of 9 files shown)


src/plugins/cordova-plugin-inappbrowser/app-host-handlers.js
```diff
@@ -0,0 +1,49 @@
...
+    return {
+        'InAppBrowser': {
+            'open': function (success, fail, args) {
+                instance = Browser.create(success, fail, args);
+            },
+            'show': function () {
+                execute('show');
+            },
+            'close': function () {
+                execute('close');
```
dlebu: Once closed, can the browser be opened again? I checked the spec but it has no documentation on this. I am wondering if we should assign the `instance` to `undefined` on close.

src/plugins/cordova-plugin-inappbrowser/inapp-browser.js
```diff
@@ -0,0 +1,391 @@
...
+IframeBrowser.prototype.hide = function () {
+    if (this._container && this._isVisible) {
+        this._container.style.display = 'none';
+        this._isVisible = false;
+    }
+};
+
+IframeBrowser.prototype.injectScriptCode = function (callback, args) {
+    if (this._container && this._frame) {
+        var code = args[0].replace(/\r?\n|\r/g, ''),
```
dlebu: Is this equivalent: `/\r?\n?/g`  ?

src/plugins/cordova-plugin-inappbrowser/app-host-handlers.js
```diff
@@ -0,0 +1,49 @@
...
+            // remove the first element since it is the "fnName" parameter
+            Array.prototype.shift.apply(arguments);
+            args = arguments;
+        }
+
+        return fn.apply(instance, args);
+    }
+
+    return {
+        'InAppBrowser': {
```
dlebu: The latest docs show a slightly different set of methods [here](https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-inappbrowser/#inappbrowser):

 
-    addEventListener
-    removeEventListener
-    close
-    show
-    executeScript
-    insertCSS

dlebu: Ignore my previous comment, I was thinking about the JS interface, while this is the exec bridge interface. 👍

src/plugins/cordova-plugin-inappbrowser/inapp-browser.js
```diff
@@ -0,0 +1,391 @@
...
+            'head.appendChild(link);' +
+            '}());';
+
+        try {
+            var result = this._injectCode(code);
+            if (hasCallback) {
+                callback(result);
+            }
+        } catch (error) {
+            console.error('Error occured while trying to inject a style file', error);
```
dlebu: Do we need to call an error callback here? Are these functions intentionally missing the error callbacks?

src/plugins/cordova-plugin-inappbrowser/app-host-handlers.js
albertinad: That is my doubt too. I didn't find anything in the spec. I will play with this plugin on a real device and see what is the behaviour. But I think that could be fine to assign undefined or null on close.

src/plugins/cordova-plugin-inappbrowser/inapp-browser.js
albertinad: This part is still incomplete, and I will add it in my next commits. I checked at the android implementation and it handles the error of the WebView and notifies.

src/plugins/cordova-plugin-inappbrowser/inapp-browser.js
```diff
@@ -0,0 +1,391 @@
...
+    case '_system':
+        // open in a new browser tab
+        Constructor = SystemBrowser;
+        break;
+    case '_blank':
+        Constructor = IframeBrowser;
+        break;
+    default:
+        // "_self" and any other option, use the same tab
+        window.location = args[0];
```
albertinad: This is wrong. The documentation of inappbrowser plugin is wrong, or I didn't understand it. But after checkign in the mobilespec app what is the expected result for `any other option` and after taking a look at the (android implementation)[https://github.com/apache/cordova-plugin-inappbrowser/blob/master/src/android/InAppBrowser.java#L188], the target `_blank` and `everything else` will be loaded in the InAppBrowser, and for us, the iframe (or a new window after I've added the configuration panel).

src/plugins/cordova-plugin-inappbrowser/inapp-browser.js
albertinad: @dlebu I tried to inject an script that throws an error, due to an invalid token for example, and any callback was executed. So this code is following at least the behaviour on an android device.
I recommend to still keep this log here, to help developers analyze why their injected code is failing.

dlebu: Sounds good.

albertinad:
I've added an option to choose the default behaviour of the InAppBrowser when the target is `_blank`: 
1 - to use the iframe implementation taking into account the security issues and CORS, and 
2 - to behave as the `_system` target, to use a new window


inappbrowser-widget

albertinad:
This has only one issue, that when navigating the iframe or even a new window to a local resource, we prepare and inject app-host for it, since we are serving every asset of the app in our server, when we shouldn't inject it. So I wanted to check with you my proposal to fix this: add a query parameter to the url, before navigating in a new window or in an iframe, and use that parameter to filter when to inject or not app-host. What do you think?

dlebu:
I think that is a good approach for this issue. We need to make that decision somehow and the query param is a good idea.

src/plugins/cordova-plugin-inappbrowser/inapp-browser.js
```diff
@@ -0,0 +1,434 @@
...
+ */
+InAppBrowser.Events = {
+    LOAD_START: 'loadstart',
+    LOAD_STOP: 'loadstop',
+    LOAD_ERROR: 'loaderror',
+    EXIT: 'exit'
+};
+
+InAppBrowser.prototype._prepareUrl = function (url) {
+    var queryParameter = 'cdvsim-simulation=false',
```
dlebu: `cdvsim-enabled`?

src/server/server.js
```diff
@@ -41,16 +41,22 @@ function sendHostJsFile(response, hostType) {
 }
 
 function streamAppHostHtml(request, response) {
-    // Always prepare before serving up app HTML file, so app is up-to-date, then create the app-host.js (if it is
-    // out-of-date) so it is ready when it is requested.
+    var filePath = path.join(config.platformRoot, url.parse(request.url).pathname);
```
dlebu: @guillaumejenkins Will this change affect the prepare fix you made in #43 ?

dlebu:
This looks good to me @albertinad . I had a parameter name proposal - not sure if it any better than the current one. Let's wait for @guillaumejenkins to take a look as well, he made prepare/livereload changes recently and I want to make sure the changes are not impacted.

src/server/server.js
guillaumejenkins: Doesn't look like it will; my changes are limited to the scope of `prepare.prepare()`

README.md
```diff
@@ -60,6 +60,7 @@ This preview version currently includes built-in support for the following Cordo
 * [cordova-plugin-file](https://github.com/apache/cordova-plugin-file)
 * [cordova-plugin-geolocation](https://github.com/apache/cordova-plugin-geolocation)
 * [cordova-plugin-globalization](https://github.com/apache/cordova-plugin-globalization)
+* [cordova-plugin-inappbrowser](https://github.com/apache/cordova-plugin-inappbrowser)
```
guillaumejenkins: Did you reuse some code from the plugin's repo? If yes, we might need to update the third-party notices

src/plugins/cordova-plugin-inappbrowser/inapp-browser.js
albertinad: That one is better! I will change it. Thanks!

albertinad:
@dlebu @guillaumejenkins I've checked and I start basing the simulation for the [`_blank` target](https://github.com/Microsoft/cordova-simulate/blob/12bd6da836a8dbd25701eba007d44a7216b1c3cd/src/plugins/cordova-plugin-inappbrowser/inapp-browser.js#L296) using the browser implementation proxy for the InAppBrowser, but it was only an starting point for [drawing the iframe](https://github.com/apache/cordova-plugin-inappbrowser/blob/master/src/browser/InAppBrowserProxy.js#L78).

PR #39, opened 2016-06-01
Removing deviceready event from sim-host.

dlebu:
As agreed, we are removing `deviceready` from sim-host.


albertinad:
Looks good to me! Thank you @dlebu!

Issue #40, opened 2016-06-02
Media simulation doesn't follow the plugin specification 


albertinad:
The current implementation of media simulation doesn't follow the plugin specification related to:
- error handling: an instance a `MediaError` object must be sent when an error happens. That [object](https://github.com/apache/cordova-plugin-media#mediaerror) contains the following properties: `code`, that seems to be required, and `message`, that it seems to be optional.
- the media instance doesn't keep track of the state, so it allows multiple calls to stop and pause APIs. For example, calling [stop](https://github.com/apache/cordova-plugin-media/blob/master/src/android/AudioPlayer.java#L340) function when the media is not active (running or paused), it should send an error with the status code `MEDIA_ERR_NON_ACTIVE` and avoid calling internal stop.

albertinad:
Fixed in #41

PR #41, opened 2016-06-02
Update media plugin simulation to be consistent with plugin specification

albertinad:
Hi! This PR is a fix for the issue #40:
- Improving error handling in order to follow the specification and send a `MediaError` object.
- Keep track of the current state, to vaildate if it is possible to perform actions such as play, pause and stop.
- Improve the creation and validation of the Audio instance.
- Minor code refactor

Thanks!

### DIFF
--- a/src/plugins/cordova-plugin-media/app-host-handlers.js
+++ b/src/plugins/cordova-plugin-media/app-host-handlers.js
@@ -87,7 +87,10 @@ module.exports = function (messages) {
                 }
             },
             'setRate': function (success, fail, args) {
-                // TODO check ios impl
+                var player = Player.getById(args[0]);
+                if (player) {
+                    player.setRate(args[1]);
+                }
             },
             'getCurrentAmplitudeAudio': function (success, fail, args) {
                 var player = Player.getById(args[0]),
--- a/src/plugins/cordova-plugin-media/app-host-handlers.js
+++ b/src/plugins/cordova-plugin-media/app-host-handlers.js
@@ -51,6 +51,11 @@ Media.MEDIA_PAUSED = 3;
 Media.MEDIA_STOPPED = 4;
 Media.MEDIA_MSG = ['None', 'Starting', 'Running', 'Paused', 'Stopped'];
 
+// "static" function to return existing objs.
+Media.get = function(id) {
+    return mediaObjects[id];
+};
+
 /**
  * Start or resume playing audio file.
  */
--- a/src/plugins/cordova-plugin-media/app-host-handlers.js
+++ b/src/plugins/cordova-plugin-media/app-host-handlers.js
@@ -7,6 +7,32 @@ var utils = require('utils'),
 
 var mediaObjects = {};
 
+function TaskQueue() {
+    this._callbacksQueue = [];
+    this._isCallbackInProgress = false;
+}
+
+TaskQueue.prototype.push = function (fn) {
+    this._callbacksQueue.push(fn);
+    this._executeNext();
+};
+
+TaskQueue.prototype._executeNext = function () {
+    if (this._callbacksQueue.length > 0 && !this._isCallbackInProgress) {
+        this._isCallbackInProgress = true;
+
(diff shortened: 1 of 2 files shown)


albertinad:
@dlebu @guillaumejenkins @busykai I have changed how the simulation of Media plugin is done. In the old implementation, it was a copy of the browser proxy implementation. After doing some tests, I find out that it doesn't behaves the same in a browser, in a real device and in the simulator. So I find some issues with that, and start fixing them in previous commits. But since we override the `window.Media` object with our simulation, and I have added some others private APIs, happens that we end up providing a different media instance for developers, with private APIs that are not available in a real device. Also we just copy most of the javascript side implementation of the real plugin for simulation, and we need to  keep in sync with the original.
In the latests commit I tried to propose a different approach:
- Do not override the `window.Media` object with something similar than the real JS interface. Just use the exactly JS that the plugin provides and write as a simulation, the native calls to exec.
- The browser implementation is not following the standard, so I based my simulation code with android native implementation
With this changes, the behaviour of the media plugin on the simulator is like in a real device.
For testing it, I use the `mobilespec app` and also a code snippet that was the driver of the changes, since with the same code base, I have really different behaviour in the simulator and in a real device, like android:
```
// add some button or element to interact with

 
function playSound() {
    var onSuccess = function() {
        media.stop();
        media.release();

        console.log("onSuccess");
    };

    var onError = function(error) {
        media.stop();
        media.release();

        console.log("onError", error);
    };

    function onStatus(status) {
        var msg;
        switch(status) {
            case Media.MEDIA_NONE: // 0
                msg = "MEDIA_NONE";
                break;
            case Media.MEDIA_STARTING: // 1
                msg = "MEDIA_STARTING";
                break;
            case Media.MEDIA_RUNNING: // 2
                msg = "MEDIA_RUNNING";
                break;
            case Media.MEDIA_PAUSED: // 3
                msg = "MEDIA_PAUSED";
                break;
            case Media.MEDIA_STOPPED: // 3
                msg = "MEDIA_STOPPED";
                break;
            default:
                msg = "unknown status";
        }
        console.log("onStatus: " + msg, status) ;
    }

    // use some short audio, a few seconds (~2 seconds)
    var src =  'audio/some-short-sound.mp3',
        media = new Media(src, onSuccess, onError, onStatus);

    media.play();
}

// at some point
document.getElementById('play-sound').addEventListener('click', playSound);
```

Now with these changes, I have the same behaviour in a real device and in the simulator.

src/plugins/cordova-plugin-media/app-host-handlers.js
```diff
@@ -2,281 +2,106 @@
...
+            'stopRecordingAudio': function (success, fail, args) {
+                var player = Player.getById(args[0]);
+                if (player) {
+                    player.stopRecord(true);
+                }
+            },
+            'pauseRecordingAudio': function (success, fail, args) {
+                var player = Player.getById(args[0]);
+                if (player) {
+                    player.stopRecord(false);
```
dlebu: Should the `pause` parameter be `true` here and `false` in the `stopRecordingAudio` call?

dlebu:
This looks good to me, @albertinad. Thanks for doing your due diligence to improve the implementation!

src/plugins/cordova-plugin-media/app-host-handlers.js
albertinad: According to the implementation [here](https://github.com/apache/cordova-plugin-media/blob/master/src/android/AudioHandler.java#L120), this should be `true` for `stopRecordingAudio` and `false` for `pauseRecordingAudio`. This is just to follow the other implementation, but when we support recording, we can change it :)

albertinad:
@dlebu No problem! :) thanks!

albertinad:
@dlebu Is it ok if I merge this?

dlebu:
@albertinad sure! :shipit:

albertinad:
Merged! thanks!

PR #43, opened 2016-06-03
Smart prepare

### LINKED ISSUE #37: The server does a cordova prepare before serving every HTML file
When requested to serve an HTML file, the server will execute a `cordova prepare`, which adds a noticeable delay when navigating within a simulated app.

Possible solutions are:

- Only prepare if the served file is the app's start page
- Only prepare if there have been changes since the last prepare

guillaumejenkins:
Fixes #37.

More context on the 2 big changes:

**Prepare**
Currently, every time an HTML file is served to app-host, the server forces a prepare. This adds between 0.5 and 1 second of delay on every navigation - including when live reloading changes.

These changes add a "project state" verification to the prepare module to prevent unnecessary prepares. The prepare module stores the list of plugins, project files and debug-host handlers and uses that to determine whether to prepare or not.

**Live reload**
In the existing live reload server, app-host must be queried every time there is a changed file to ask it whether that file can be refreshed without a full page reload. The reason it had to do that was to prevent preparing twice in a row in case the app required a full reload and the user had specified `--forceprepare`.

With these changes to the prepare module, it no longer matters if the live reload server requests an extra prepare - the prepare module is now smart enough to not prepare if it is not needed. So, that 3-way messenging between the live reload server and app-host can be entirely removed.

### DIFF
--- a/src/app-host/live-reload-client.js
+++ b/src/app-host/live-reload-client.js
@@ -3,12 +3,6 @@
 
 var url = require('url');
 
-var liveReloadEvents = {
-    CAN_REFRESH: 'lr-can-refresh',
-    REFRESH_FILE: 'lr-refresh-file',
-    FULL_RELOAD: 'lr-full-reload'
-};
-
 var URL_ATTRIB_NAME = 'url';
 var HREF_ATTRIB_NAME = 'href';
 var SRC_ATTRIB_NAME = 'src';
@@ -26,18 +20,11 @@ module.exports.start = function (sock) {
         '/',
         ''
     ];
-    var pendingFilesToRefresh = {};
     var socket = sock;
 
     /**
-     * Reload the page. Currently, only does a naive window.location.reload().
-     */
-    function reloadPage() {
-        window.location.reload(true);
-    }
-
-    /**
-     * Returns the name of the reference attribute (either "url", "href" or "src") that is defined for the given node. If the node defines more than one, returns the first encountered, in that order.
+     * Returns the name of the reference attribute (either "url", "href" or "src") that is defined for the given node.
+     * If the node defines more than one, returns the first encountered, in that order.
      *
      * @param {Element} domNode The DOM node to check.
      * @returns {String} "url", "href" or "src", or null if none of these attributes is defined.
@@ -78,13 +65,15 @@ module.exports.start = function (sock) {
     }
 
     /**
-     * Finds all the DOM elements that have a reference attribute ("url", "href" or "src") pointing to the given relative path. Excludes <script> tags.
+     * Finds all the DOM elements that have a reference attribute ("url", "href" or "src") pointing to the given
+     * relative path. Excludes <script> tags.
      *
      * @param {String} fileRelativePath The URL of the file to check, relative to the webRoot.
      * @returns {{ domNode: Element, referenceAttribute: string }[]} An array of "resources" referencing the given file.
      */
     function findDomNodesForFilePath(fileRelativePath) {
-        // To use querySelectorAll to query elements based on their attributes, the selector's syntax is: '[attrib1], [attrib2], ...'.
+        // To use querySelectorAll to query elements based on their attributes, the selector's syntax is:
+        // '[attrib1], [attrib2], ...'.
         var selectorString = `[${referenceAttributes.join('], [')}]`;
         var rawNodes = document.querySelectorAll(selectorString);
         var filteredNodes = [];
@@ -118,20 +107,21 @@ module.exports.start = function (sock) {
     }
 
     /**
-     * Determines whether a file can be refreshed without a full page reload, and notifies the server. If the file can be refreshed, stores the relevant info in the list of pending files to refresh.
+     * Determines whether the changes form the specified file can be applied to the app without a full page reload.
+     * Then, based on that, either updates the reference attribute of the appropriate node, or does a full page
+     * reload. 
      *
-     * @param {String} fileRelativePath The URL of the file to be refreshed, relative to the webRoot.
+     * @param {String} fileRelativePath The URL of the file that changed, relative to the webRoot.
      */
-    function prepareRefresh(fileRelativePath) {
+    function onFileChanged(fileRelativePath) {
         var associatedNodes = findDomNodesForFilePath(fileRelativePath);
         var canRefresh = false;
 
         if (associatedNodes.length) {
-            pendingFilesToRefresh[fileRelativePath] = associatedNodes;
-            canRefresh = true;
+            refreshFile(fileRelativePath, associatedNodes);
+        } else {
+            reloadPage();
         }
-
-        socket.emit(liveReloadEvents.CAN_REFRESH, { fileRelativePath: fileRelativePath, canRefresh: canRefresh });
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/app-host/live-reload-client.js
```diff
@@ -118,37 +107,30 @@ module.exports.start = function (sock) {
...
-     * Determines whether a file can be refreshed without a full page reload, and notifies the server. If the file can be refreshed, stores the relevant info in the list of pending files to refresh.
+     * Determines whether the changes form the specified file can be applied to the app without a full page reload.
+     * Then, based on that, either updates the reference attribute of the appropriate node, or does a full page
+     * reload. 
      *
-     * @param {String} fileRelativePath The URL of the file to be refreshed, relative to the webRoot.
+     * @param {String} fileRelativePath The URL of the file that changed, relative to the webRoot.
      */
-    function prepareRefresh(fileRelativePath) {
+    function fileChanged(fileRelativePath) {
```
dlebu: onFileChanged?

src/server/prepare.js
```diff
@@ -1,30 +1,54 @@
...
+        if (shouldPrepare(currentState)) {
+            execPromise = execCordovaPrepare().then(() => true);
+        } else {
+            execPromise = Q(false);
+        }
+
+        execPromise.then((didPrepare) => {
+            if (didPrepare || shouldInitPlugins(currentState)) {
+                plugins.initPlugins();
+            } else {
```
dlebu: Do we need this else clause?

src/server/prepare.js
```diff
@@ -68,6 +86,110 @@ function execCordovaPrepare() {
...
+    if (!previousState) {
+        return true;
+    }
+
+    currentState = currentState || getProjectState();
+
+    // Use JSON stringification to compare states. This works because a state is an object that only contains
+    // serializable values, and because the glob module, which is used to read directories recursively, is
+    // deterministic in the order of the files it returns.
+    var pluginsAreTheSame = JSON.stringify(currentState.pluginList) === JSON.stringify(previousState.pluginList);
```
dlebu: We should not rely on glob's deterministic order here.

src/server/prepare.js
```diff
@@ -68,6 +86,110 @@ function execCordovaPrepare() {
...
```
dlebu: Same here. Let's create a utility method that compares two lists and use it for all comparisons.

src/server/prepare.js
```diff
@@ -68,6 +86,110 @@ function execCordovaPrepare() {
...
+function getProjectState() {
+    var currentPlatform = config.platform;
+    var newState = {};
+
+    // Get the list of plugins for the current platform.
+    var pluginsJsonPath = path.join(config.projectRoot, 'plugins', currentPlatform + '.json');
+    var installedPlugins = {};
+
+    try {
+        var fileContent = fs.readFileSync(pluginsJsonPath).toString();
```
dlebu: If possible, we should keep all interactions with the file system asynchronous.

src/app-host/live-reload-client.js
guillaumejenkins: Sure

src/server/prepare.js
guillaumejenkins: Nope :)

src/server/prepare.js
guillaumejenkins: sim-files had a method like that, so I moved it to a new utils module and now use that

src/server/prepare.js
guillaumejenkins: Converted everything to async

src/modules/jsUtils.js
```diff
@@ -0,0 +1,28 @@
...
+    if (typeof o1 === 'undefined' || typeof o1 === 'undefined') {
+        return false;
+    }
+
+    if (Array.isArray(o1)) {
+        if (!Array.isArray(o2)) {
+            return false;
+        }
+        // Simple array comparison - expects same order and only scalar values
+        return o1.length === o2.length && o1.every(function (v, i) {
```
dlebu: Do we care about the scenario when the order of elements is different? For our scenario, are two arrays with the same elements in different order not equal?

guillaumejenkins: I greatly improved the object comparison method. It now supports nested objects and supports comparing order or not.

src/modules/jsUtils.js
```diff
@@ -0,0 +1,94 @@
+/**
```
albertinad: We have a `utils` module already, that has this kind of utility functions. Maybe we can place this in there, if we keep growing that file, then we should separate it, but right now, it only has js utils and some location utilities. It is not too big.

guillaumejenkins: I had initially placed these functions in `utils.js`, but I had runtime errors: `Error: Cannot find module 'exception'`. Because there is no `exception` dependency in our `package.json`, I assume this means that the current `utils` module was meant to be run in a `browserify`'d environment, which is not the case for the server code.

Because of that, I thought the best approach was to have a separate `jsUtils` (or would `serverUtils` be better?) module that would not require to be inside a `browserify`'d environment to work.

@albertinad @dlebu what do you think? Would there be an alternative to having 2 separate util modules (one for `browserify`, and one for pure standalone JS)?

src/modules/jsUtils.js
```diff
@@ -0,0 +1,94 @@
...
+ * protection against circular objects. By default, order of object properties and array values is not important. 
+ * 
+ * @param {any} o1 The first object.
+ * @param {any} o2 The second object.
+ * @param {boolean=} compareOrder Whether order in object keys and array values should be considered.
+ * 
+ * @returns {boolean} Whether the objects are similar.
+ */
+function compareObjects(o1, o2, compareOrder) {
+    // If either are undefined, return false - don't consider undefined to equal undefined.
```
dlebu: We could make the comparison faster if we compare the objects first.

```javascript
if (o1 === o2) {
   return true;
} else {
 // continue with current approach
}

```

src/modules/jsUtils.js
```diff
@@ -0,0 +1,94 @@
...
+ * default, order of values is not important ([1,2,3] and [3,2,1] are considered similar).
+ * 
+ * @param {any[]} a1 The first array.
+ * @param {any[]} a2 The second array.
+ * @param {boolean=} compareOrder Whether order of values should be considered.
+ * 
+ * @returns {boolean} Whether the arrays are similar.
+ */
+function compareArrays(a1, a2, compareOrder) {
+    // If either are undefined, return false - don't consider undefined to equal undefined.
```
dlebu: Same here.

src/server/prepare.js
```diff
@@ -1,30 +1,53 @@
...
 function prepare() {
     if (!preparePromise) {
         var d = Q.defer();
+        var currentProjectState;
 
         preparePromise = d.promise;
         lastPlatform = config.platform;
 
-        execCordovaPrepare().finally(function () {
+        Q().then(() => {
```
dlebu: You could `return getProjectState().then(...)` directly here.

dlebu:
Looks good to me with some comments.

src/app-host/live-reload-client.js
```diff
@@ -78,13 +65,15 @@ module.exports.start = function (sock) {
...
+     * relative path. Excludes <script> tags.
      *
      * @param {String} fileRelativePath The URL of the file to check, relative to the webRoot.
      * @returns {{ domNode: Element, referenceAttribute: string }[]} An array of "resources" referencing the given file.
      */
     function findDomNodesForFilePath(fileRelativePath) {
-        // To use querySelectorAll to query elements based on their attributes, the selector's syntax is: '[attrib1], [attrib2], ...'.
+        // To use querySelectorAll to query elements based on their attributes, the selector's syntax is:
+        // '[attrib1], [attrib2], ...'.
         var selectorString = `[${referenceAttributes.join('], [')}]`;
```
albertinad: Templates is a feature of ecmascript 6, I know that we are targeting like new browsers, but maybe we can consider using features of the language and APIs that are most common in other browsers too until we have full support of it. I don't know the state of the support of templates, so this is just a comment :)

src/modules/jsUtils.js
albertinad: I see, right, you have an error since you are in node context. I think that is OK to have a separate module that the server is free to uses. But I would suggest to put that module in other location, since that could be confusing. One can think that if the server can require this module, under `src/modules/` then any other server thing can be placed in here, when it is not alwasy true. The intention of the modules that are in this folder I think that it is for browser usages, by using browserify.
Maybe we can improve a little how the `utils` module is implemented, in order to separate functions that are more useful for simulation, like navigation utils for example, from functions that we can have in a kind of utility library itself. Then we can point in the browserify configuration, to load that module to.

albertinad:
Looks good to me!

PR #44, opened 2016-06-06
Moved geolocation to first position in sim-host.

### LINKED ISSUE #9: Geolocation panel: map image disappears when panel is stacked horizontally
Tested in Chrome. When the plugin UI window is narrow and the Geolocation panel is stacked vertically with the other panels, everything is fine:

image

However, when you resize the page to be wider and the Geolocation panel is moved to the right of other panels, the map background disappears:

image

dlebu:
Fixes #9.


albertinad:
This looks good to me.

PR #46, opened 2016-06-06
Commenting out unsupported camera options

guillaumejenkins:
Commenting out code related to the currently unsupported camera options (host system camera and built-in default image).

The reason why I don't want to delete the code is that we will very likely re-enable these options soon, and the commented code will provide an easier starting point.

Tracking for this has been opened in #47.


albertinad:
This looks good to me.

dlebu:
LGTM!

PR #48, opened 2016-06-07
Replacing textContent property with value property name.

albertinad:
Hi! This PR is a workaround proposal for an issue that I'm having with the `cordova-label` web component on old versions of chromium: the `textContent` property of the component is not re-defined so we never end up modifying the  `label` element inside the `shadowRoot`.


dlebu:
This looks good to me.

PR #49, opened 2016-06-07
Removing usage of string templates

guillaumejenkins:
They are unnecessary and break support for several older browser.


albertinad:
Thank you @guillaumejenkins ! LGTM!

dlebu:
LGTM!

PR #50, opened 2016-06-07
Update package.json

dlebu:
Version change for 0.2.0 release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cordova-simulate",
-  "version": "0.2.0-dev",
+  "version": "0.2.0",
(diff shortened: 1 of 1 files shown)


albertinad:
Looks good to me!
I noticed that we are using [lowercase](https://github.com/Microsoft/cordova-simulate/blob/f977b57df62b264a2fbbada2cb7628f0f1593085/package.json#L12) for the whole endpoint to GitHub repository, it redirects too, but should be use `Microsoft` instead of `microsoft`, as the repo is?

dlebu:
I changed it to uppercase. Thanks @albertinad !

PR #51, opened 2016-06-08
Allowing project reinit.

dlebu:
I made a change in our integration code and I started hitting the "Cannot reinitialize..." exception. Given that we can close and restart the simulation server programmatically via the API, I think we should allow project path re-initialization.

### DIFF
--- a/src/server/config.js
+++ b/src/server/config.js
@@ -13,7 +13,7 @@ var simulationFilePath;
     { name: 'liveReload' },
     { name: 'platform' },
     { name: 'platformRoot' },
-    { name: 'projectRoot', single: true },
+    { name: 'projectRoot' },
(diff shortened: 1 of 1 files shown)


albertinad:
@dlebu In our integration code, we need an `stopSimulation` API in order to relaunch a new simulation server with a fresh configuration without spawning a new process or unload modules, this could help you too. I didn't contribute this because I wanted to clean it up a little more, propose a new design, but before doing that, it would be better that you take a look at it too #52

dlebu:
@albertinad Thanks! Given that #52 re-initializes the config object, I'm inclined to close this PR. I don't think there will be valid scenarios in which we would like to reinit the project root without calling stop simulate first.

PR #52, opened 2016-06-08
[review] Stop simulation programmatically API

albertinad:
Hi! This PR is to show some changes I made in order to be able to programmatically stop the simulation, and start a new one, with the same project and configuration or with different one.
I would like to share this with you, in order to see some of the changes that we need and to start the discussion about the right approach to follow.
Thanks!

### DIFF
--- a/src/server/config.js
+++ b/src/server/config.js
@@ -4,7 +4,6 @@ var fs = require('fs'),
     path = require('path');
 
 var config = {};
-var simulationFilePath;
 
 // module properties
 [
@@ -47,7 +46,6 @@ function getValue(prop, optional) {
 
 function newInstance() {
     config = {};
-    simulationFilePath = null;
 }
 
 module.exports.newInstance = newInstance;
--- a/src/server/config.js
+++ b/src/server/config.js
@@ -44,3 +44,10 @@ function getValue(prop, optional) {
     }
     return config[prop];
 }
+
+function newInstance() {
+    config = {};
+    simulationFilePath = null;
+}
+
+module.exports.newInstance = newInstance;
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -130,6 +130,21 @@ function processPluginHtml(html, pluginId) {
     });
 }
 
+function stop() {
+    plugins.clearPlugins();
+    simFiles.clearSimFiles();
+    config.newInstance();
+
+    if (config.server) {
+        config.server.close(function (error) {
+            if (error) {
+                log.log(error);
+            }
+        })
+    }
+}
+
 module.exports = {
-    attach: attach
+    attach: attach,
+    stop: stop
 };
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -135,9 +135,9 @@ function stop() {
     simSocket.closeConnections();
 
     config.newInstance();
-    plugins.clear();
-    simFiles.clear();
-    prepare.clear();
+    plugins.reset();
+    simFiles.reset();
+    prepare.reset();
 }
 
 module.exports = {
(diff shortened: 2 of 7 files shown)


src/server/config.js
```diff
@@ -58,3 +58,10 @@ function getValue(prop, optional) {
     }
     return config[prop];
 }
+
+function newInstance() {
+    config = {};
+    simulationFilePath = null;
```
dlebu: This is part of the config now. I think you still need to merge master into your branch.

albertinad: Yes, this branch is out of date, for a week I guess, I will update it today :) in a couple of minutes. Thanks @dlebu !

src/server/server.js
```diff
@@ -124,6 +125,16 @@ function processPluginHtml(html, pluginId) {
     });
 }
 
+function stop() {
```
dlebu: We should add this function as an export to `src/simulate.js` as well to expose it to npm.

dlebu: I opened a PR for this: https://github.com/albertinad/taco-simulate-server/pull/1

src/server/server.js
```diff
@@ -130,6 +131,16 @@ function processPluginHtml(html, pluginId) {
...
+    simSocket.closeConnections();
+
+    config.newInstance();
+    plugins.clear();
+    simFiles.clear();
+    prepare.clear();
```
busykai: Maybe `reset` is a more indicative name for these functions (as opposed to `clear`).

dlebu:
This looks good to me - please merge the linked PR in first. Thanks for the fix!

src/server/server.js
albertinad: You're right, I will rename them to `reset`. Thanks @busykai!

albertinad:
Using this as it is, after stop the simulation for project A, and start a new one for project B, I noticed the following error in some of the hosts page:
```
WebSocket connection to 'ws://localhost:8000/socket.io/?EIO=3&transport=websocket&sid=d0GMqTmyyah69ktlAAAG' failed: Connection closed before receiving a handshake response
```

Calling `server.close()` stops the server receiving new connections, but keeps the current connections, and it is closed when all active connections are closed. So, what I did was to add this function that is called when the server has been created: 
```
function trackServerConnections(srv) {
    var nextSocketId = 0;
    connections = {};
    srv.on('connection', function (socket) {
        var socketId = nextSocketId++;
        connections[socketId] = socket;
        console.log('socket', socketId, 'opened');

        // Remove the socket when it closes
        socket.on('close', function () {
            console.log('socket', socketId, 'closed');
            delete connections[socketId];
        });
    });
}
```

And now the close and stopSimulation functions looks like this,
```
var closeServer = function () {
    server.server && server.server.close();

    for (var socketId in connections) {
        console.log('socket', socketId, 'destroyed');
        connections[socketId].destroy();
    }
};

var stopSimulate = function () {
    closeServer();
    simServer.stop();
    server = null;
};
```

And with these changes it looks like it works fine!

src/simulate.js
```diff
@@ -9,11 +9,13 @@ var fs = require('fs'),
     simSocket = require('./server/socket'),
     dirs = require('./server/dirs');
 
-var server = cordovaServe();
+var server;
 
 var launchServer = function (opts) {
     opts = opts || {};
 
+    server = cordovaServe();
```
albertinad: We should create a new CordovaServe instance, that internally creates a new express app for us.

src/server/socket.js
```diff
@@ -152,6 +163,27 @@ function invalidateSimHost() {
     hostSockets[SIM_HOST] = null;
 }
 
+function closeConnections() {
+  var hostType,
+      socket;
+  for (hostType in hostSockets) {
+    if (hostSockets.hasOwnProperty(hostType)) {
```
guillaumejenkins: Small nit: by using `Object.keys(hostSockets).forEach()`, we don't need to check for `hasOwnProperty()`

PR #54, opened 2016-06-08
Basic linting

busykai:
0. Base infrastructure on top of `gulp` and `eslint`.
1. Basic recommended checks are enabled.
2. Major issues fixed (except for ref to `Windows` in network-information).
3. Third-party code is ignored.
4. Logging on `console` is allowed anywhere except for `src/app-host` (which is incorrect since both `src/platforms` and `src/plugins` contain a lot of code executed on `APP_HOST`).

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,7 +2,6 @@
     "extends": "eslint:recommended",
     "env": {
         "browser": true,
-        "es6": true,
         "commonjs": true
     },
     "rules": {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,12 @@
+{
+    "extends": "eslint:recommended",
+    "env": {
+        "browser": true,
+        "es6": true,
+        "commonjs": true
+    },
+    "rules": {
+        "quotes": [ 2, "single", { "avoidEscape": false, "allowTemplateLiterals": true } ],
+        "no-console": [ 0 ]
+    }
+}
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,11 +1,21 @@
-var gulp = require('gulp'),
-    eslint = require('gulp-eslint');
+var gulp        = require('gulp'),
+    eslint      = require('gulp-eslint'),
+    jsonlint    = require('gulp-jsonlint');
 
-gulp.task('lint', function () {
+gulp.task('eslint', function () {
     return gulp.src('src/**/*.js')
         .pipe(eslint())
-        .pipe(eslint.format());
-        //.pipe(eslint.failOnError());
+        .pipe(eslint.format())
+        .pipe(eslint.failAfterError());
 });
+
+gulp.task('jsonlint', function () {
+    return gulp.src([ 'src/**/*.json', 'package.json' ])
+        .pipe(jsonlint())
+        .pipe(jsonlint.reporter())
+        .pipe(jsonlint.failAfterError());
+});
+
+gulp.task('lint', [ 'eslint', 'jsonlint' ]);
 
 gulp.task('default', [ 'lint' ]);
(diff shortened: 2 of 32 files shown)


src/app-host/app-host.js
```diff
@@ -1,11 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
-
+/* global io:false */
 var livereload = require('./live-reload-client');
 var Messages = require('messages');
 var telemetry = require('telemetry-helper');
 
 var cordova;
-var oldExec;
 var socket = io();
```
albertinad: what about using io as `window.io`, and it is clear that is a global window variable, in the context of the browser? then we don't need to ignore the rule.

src/modules/sim-constants.js
```diff
@@ -2,213 +2,213 @@
 // Based in part on code from Apache Ripple, https://github.com/apache/incubator-ripple
```
albertinad: These constants comes from Ripple, and we use them in our built-in plugins simulation. But do we really want all of them? Some of them could be defined only as part of the plugin simulation, and not need to be defined here. There's no need to change anything related to this as part of this PR, I just bring the question :)

src/app-host/app-host.js
busykai: but we don't ignore the rule, we just tell that there's a global `io` which is not writeable (see line 2). the code does not need to become more verbose just because it's being linted.

src/modules/sim-constants.js
busykai: yes, i looked at it and we certainly do not need all of them. however, just like you say, one issue at a time. feel free to file a bug for this specific file. there's actually quite a lot of clean up required through out the code, but i think it could be done over time.

src/app-host/app-host.js
albertinad: You're right @busykai, it totally makes sense.

src/modules/sim-constants.js
albertinad: Sure! I already create an issue to address this clean up, thanks @busykai

albertinad:
These changes looks good to me! thank you @busykai

.eslintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+    "extends": "eslint:recommended",
+    "env": {
+        "browser": true,
+        "es6": true,
```
dlebu: Is this accurate? #56 addresses ES6 usage to preserve backward compatibility. @albertinad

gulpfile.js
```diff
@@ -0,0 +1,11 @@
+var gulp = require('gulp'),
+    eslint = require('gulp-eslint');
+
+gulp.task('lint', function () {
+    return gulp.src('src/**/*.js')
```
dlebu: Is it possible to enable linting for .json files as well?

dlebu:
I had a couple comments. Other than that, this looks good to me!

.eslintrc.json
busykai: Actually, maybe not. ES6 arrow functions were there, but I didn't check it after a rebase on top of the latest `master`.

busykai: depending on what's merged first, we will remove it in either this PR or the other. otherwise it will not be a clean run (it has only 1 failure right now).

gulpfile.js
busykai: good idea, will update the PR.

.eslintrc.json
albertinad: @busykai I'm about to merge the #56, then you can rebase with master.

albertinad: @busykai @dlebu we are ready now!

busykai:
Done. Added JSON linter. One remaining lint failure is worked around (with a comment and a reference to the issue).

albertinad:
Looks good to me!

busykai:
Merging.

Issue #55, opened 2016-06-08
Remove ES6 syntax features to ensure backward compatibility with node.


dlebu:

albertinad:
Fixed in #56

PR #56, opened 2016-06-09
Updating usages of ES6 features for backward compatibility with older node versions

albertinad:
This PR includes:
- Updating usages of arrow functions to regular ES5 functions
- Replacing usages of `Array.prototype.find` with `Array.prototype.some`

An extra: updating implementation of `server/socket` `closeConnections` function as suggested in previous PR.

This address #55 

Thanks!

### DIFF
--- a/src/app-host/live-reload-client.js
+++ b/src/app-host/live-reload-client.js
@@ -54,7 +54,7 @@ module.exports.start = function (sock) {
      */
     function urlMatchesPath(url, fileRelativePath) {
         return localUrlPrefixes.some(function (prefix) {
-            return (prefix + fileRelativePath === url);
+            return (prefix + fileRelativePath) === url;
         });
     }
 
@@ -109,7 +109,6 @@ module.exports.start = function (sock) {
      */
     function onFileChanged(fileRelativePath) {
         var associatedNodes = findDomNodesForFilePath(fileRelativePath);
-        var canRefresh = false;
 
         if (associatedNodes.length) {
             refreshFile(fileRelativePath, associatedNodes);
--- a/src/app-host/live-reload-client.js
+++ b/src/app-host/live-reload-client.js
@@ -53,15 +53,9 @@ module.exports.start = function (sock) {
      * @returns {boolean} Whether the URL points to the modified file from the server.
      */
     function urlMatchesPath(url, fileRelativePath) {
-        var isMatch = false;
-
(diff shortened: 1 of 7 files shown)


src/server/socket.js
```diff
@@ -164,23 +164,19 @@ function invalidateSimHost() {
...
-      socket;
-  for (hostType in hostSockets) {
-    if (hostSockets.hasOwnProperty(hostType)) {
-      socket = hostSockets[hostType];
-      if (socket) {
-        socket.disconnect(true);
-      }
-    }
-  }
+    Object.keys(hostSockets).forEach(function (hostType) {
```
albertinad: Thanks @guillaumejenkins for your suggestion in the previous PR!

src/server/jsUtils.js
```diff
@@ -2,17 +2,17 @@
...
  * @param {boolean=} compareOrder Whether order in object keys and array values should be considered.
- * 
+ *
  * @returns {boolean} Whether the objects are similar.
  */
 function compareObjects(o1, o2, compareOrder) {
     // If either are undefined, return false - don't consider undefined to equal undefined.
-    if (typeof o1 === 'undefined' || typeof o1 === 'undefined') {
+    if (typeof o1 === 'undefined' || typeof o2 === 'undefined') {
         return false;
```
albertinad: I've changed this condition here too, maybe we want to check for `o2` too.

dlebu: Good catch!

dlebu:
This looks good to me!

src/app-host/live-reload-client.js
```diff
@@ -53,15 +53,9 @@ module.exports.start = function (sock) {
...
      */
     function urlMatchesPath(url, fileRelativePath) {
-        var isMatch = false;
-
-        localUrlPrefixes.find(function (prefix) {
-            isMatch = prefix + fileRelativePath === url;
-
-            return isMatch;
+        return localUrlPrefixes.some(function (prefix) {
+            return (prefix + fileRelativePath === url);
```
busykai: these parenthesis do not make much sense. they are either not needed or, in case the precedence needs to be emphasized, the concatenation should be within the parenthesis, like this: `return (prefix + fileRelativePath) === url`.

albertinad: I will change it too, good catch!

PR #58, opened 2016-06-09
Updated release notes and license file.

dlebu:


RELEASENOTES.md
```diff
@@ -1,5 +1,10 @@
 # cordova-simulate Release Notes
 
+### 0.2.0 (June 9, 2016)
+* Support for additional core plugins
+* Live reload
+* Multiple bug fixes and improvements
```
guillaumejenkins: Should we list which new plugins we support?

Should we add "Support for changing the simulated screen size"? I know we implemented it as a "plugin", but I don'T think it fits in the 1st item :)

RELEASENOTES.md
```diff
@@ -1,5 +1,13 @@
 # cordova-simulate Release Notes
 
+### 0.2.0 (June 9, 2016)
+* Support for additional core plugins: cordova-plugin-battery-status, cordova-plugin-network-information, cordova-plugin-device-orientation, cordova-plugin-inappbrowser
```
dlebu: @albertinad Did I miss any new plugin here?

albertinad: That's all the new ones!

busykai:
@dlebu, I believe the `0.2.0` tag needs to be moved to this commit merging this PR. Perhaps other PRs should be stalled until `0.2.0` is properly documented and tagged.

Also, right now it does not seem to be a problem (since we are few), but later we may want to be doing this from a release branch.

dlebu:
@busykai I agree, once I merge this in, I will update the tag.

Issue #70, opened 2016-06-09
[Bug] Geolocation.getPermission is not being handled automatically


digeff:
While testing this project: https://github.com/ryanjsalva/cordova-simulate-tests/
Using the Geolocation/Locate button with 
Windows 10 with Simulate Android
cordova -v 5.4.1
cordova platforms: android 4.1.1, ios 3.9.2

cordova plugins:
cordova-plugin-add-swift-support 1.1.0 "AddSwiftSupport"
cordova-plugin-battery-status 1.1.2 "Battery"
cordova-plugin-camera 2.2.0 "Camera"
cordova-plugin-compat 1.0.0 "Compat"
cordova-plugin-contacts 2.1.0 "Contacts"
cordova-plugin-dbmeter 1.0.3 "DBMeter"
cordova-plugin-device 1.1.2 "Device"
cordova-plugin-device-motion 1.2.1 "Device Motion"
cordova-plugin-device-orientation 1.0.3 "Device Orientation"
cordova-plugin-dialogs 1.2.1 "Notification"
cordova-plugin-file 4.2.0 "File"
cordova-plugin-geolocation 2.2.0 "Geolocation"
cordova-plugin-globalization 1.0.3 "Globalization"
cordova-plugin-inappbrowser 1.4.0 "InAppBrowser"
cordova-plugin-media-capture 1.3.0 "Capture"
cordova-plugin-network-information 1.2.1 "Network Information"
cordova-plugin-splashscreen 3.2.2 "Splashscreen"
cordova-plugin-statusbar 2.1.3 "StatusBar"
cordova-plugin-vibration 2.1.1 "Vibration"
cordova-plugin-whitelist 1.2.2 "Whitelist"
cordova-sqlite-storage 1.4.1 "Cordova sqlite storage plugin"

I get an UNHANDLED EXEC CALL asking for Geolocation.getPermission.:
![image](https://cloud.githubusercontent.com/assets/14098140/15947076/930a6610-2e4e-11e6-94ea-2e5d9826f895.png)

dlebu:
This reproduces on OSX as well, but only for the Android platform.

albertinad:
@dlebu @guillaumejenkins Can I take this issue? This should happen only on android, since this exec call is called on android only. Usually in the device, the developer will be ask to grant permissions to the sensor, but in this case, we can default always to true.

dlebu:
Sure!

albertinad:
Fixed in #75

Issue #62, opened 2016-06-09
[bug][geolocation] Zoom level with wheel goes from 1 to 18 but with buttons from 0 to 21


digeff:
I can only get to Zoom Levels: 1 to 18 by using the mouse wheel, but if I use the mouse buttons I can get to 0 and 21 also.

albertinad:
@dlebu What do you think if we also change the (+) (-) buttons to a slider? That way we keep consistency with the other usages too.

albertinad:
Fixed in #75

PR #75, opened 2016-06-10
Geolocation simulation widget changes

albertinad:
Hi! This PR includes the following:
- Move the map container to render first than the inputs to update the longitude, latitude etc values. The motivation was that maybe the users are more likely to interact more with the map itself, and advanced users or those who has specific coordinates, will use then the inputs. I'm open to suggestions :)
- Implement the `Geolocation.getPermission` exec handler: always executes the success callback. Fix for #70 
- Create a new web component, `cordova-number-entry` that wraps an input type number.
- Update the geolocation inputs fields to use the `cordova-number-entry` for the values that must be numbers.
- Fix the right maximum zoom value supported by OpenLayers, it seems to be 18, otherwise a greater number will not update the map zoom. Fix for #62 

screen shot 2016-06-10 at 2 17 17 am

cc @guillaumejenkins @dlebu @busykai 

Thanks!

### DIFF
--- a/src/sim-host/ui/custom-elements.js
+++ b/src/sim-host/ui/custom-elements.js
@@ -152,6 +152,45 @@ function initialize() {
         this.classList.add('cordova-group');
     }, 'input');
 
+    registerCustomElement('cordova-number-entry', {
+        value: {
+            set: function (value) {
+                setValueSafely(this.shadowRoot.querySelector('input'), 'value', value);
+            },
+
+            get: function () {
+                return this.shadowRoot.querySelector('input').value;
+            }
+        },
+        disabled: {
+            set: function (value) {
+                setValueSafely(this.shadowRoot.querySelector('input'), 'disabled', value);
+            }
+        }
+    }, function () {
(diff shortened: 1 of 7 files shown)


dlebu:
This looks good to me.

src/sim-host/ui/custom-elements.js
```diff
@@ -152,6 +152,45 @@ function initialize() {
...
+            }
+        },
+        disabled: {
+            set: function (value) {
+                setValueSafely(this.shadowRoot.querySelector('input'), 'disabled', value);
+            }
+        }
+    }, function () {
+        this.shadowRoot.querySelector('label').textContent = this.getAttribute('label');
+        this.classList.add('cordova-panel-row');
```
guillaumejenkins: `classList` is only supported by IE 10+. Are we fine with that?

albertinad: No problem!

albertinad:
Merged! thanks!

PR #79, opened 2016-06-13
Fix several issues with the viewport panel

### LINKED ISSUE #59: Suggestion: Show (px) for Width and Height in Viewport (Screen Size)
Instead of:
Width: 480
Height: 800

Show:
Width (px): 480
Height (px): 800

### LINKED ISSUE #60: Sometimes the Viewport option selected doesn't get applied
While changing several times between Use full browser window and Use predefined screen size, sometimes the selection is not correctly applied.

### LINKED ISSUE #67: [Suggestion] Viewport panel: remember which device was selected across sessions
If I'm using the predefined screen size of iPad, it'd be nice if the simulate remembered that option the next time I use it.

### LINKED ISSUE #78: Wrong default device in viewport and device panels
According to the code, the default selected device in the list of both the viewport and device panels should be the "Generic WVGA". However, the first one in the list ("Acer A500") is selected instead.

guillaumejenkins:
Fixes #59
Fixes #60
Fixes #67
Fixes #78

### DIFF
--- a/src/plugins/cordova-plugin-device/sim-host.js
+++ b/src/plugins/cordova-plugin-device/sim-host.js
@@ -378,15 +378,32 @@ function initialize() {
         deviceList.appendChild(option);
     });
 
-    deviceList.addEventListener('change', handleSelectDevice);
-    deviceList.value = 'WVGA';
-    handleSelectDevice();
+    var defaultDeviceId = 'WVGA';
+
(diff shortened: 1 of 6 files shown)


dlebu:
This looks good to me.

dlebu:
Don't forget to run the the new linter via gulp before checking in.

PR #81, opened 2016-06-13
Fixed exec panel breaking when deleting the 'No values saved' item

### LINKED ISSUE #73: [Bug] deleting the "No values saved" from under Persisted Exec Calls breaks the widget view
1. Delete the "No values saved" from under Persisted Exec Calls by pressing the X.
2. Add a new persisted exec call. It doesn't appear on the list.

guillaumejenkins:
Fixes #73

Edit: I changed the "No values saved" text from being a list item to being a label that we show or hide when appropriate.

===
**(The following is no longer relevant)**
A better fix would have been to change "No values saved" from a list item to a label that we just hide / show when appropriate, but unfortunately, without the use of the `hidden` attribute or the `classList` property, both of which are HTML 5, it was becoming too complex for something of such low importance.

Instead, this fix just re-adds the "No values saved" item if it is deleted, and prevents trying to remove "No values saved" from the list of saved sim results if that's the item that was deleted.

### DIFF
--- a/src/plugins/exec/sim-host.js
+++ b/src/plugins/exec/sim-host.js
@@ -6,18 +6,25 @@
 var savedSims = require('./saved-sims');
 var event = require('event');
 
+var emptyLabel;
+var execList;
+
 module.exports = {
     initialize: function () {
         var sims = savedSims.sims;
 
-        var execList = document.getElementById('exec-list');
+        emptyLabel = document.getElementById('empty-label');
+        execList = document.getElementById('exec-list');
         execList.addEventListener('itemremoved', function (e) {
             savedSims.removeSim(e.detail.itemIndex);
-            addEmptyItem();
+
+            if (!savedSims.sims.length) {
+                showEmptyLabel();
+            }
         });
 
         event.on('saved-sim-added', function (sim) {
-            removeEmptyItem();
+            hideEmptyLabel();
             execList.addItem(cordovaItemFromSim(sim));
         });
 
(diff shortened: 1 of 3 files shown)


dlebu:
This looks good to me.

TimBarham:
Glad you fixed this :). But seriously what browser are you wanting to support that doesn't support `classList`? My plan had always been to just not show the 'x' for that item.

BTW, if I recall, this list grows indefinitely - is that still the case?

albertinad:
I don't think that using `classList` is wrong. It has a wide support and compatibility according [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList), and we are making use of it already. If we plan to support browsers which don't has a wide support of HTML5 features, then we should consider using some kind of polyfills or selector libraries like jQuery or similar.

src/plugins/exec/sim-host.js
```diff
@@ -51,36 +54,13 @@ function cordovaItemFromSim(sim) {
...
-    if (!hasEmptyItem) {
-        return;
-    }
-
-    var execList = document.getElementById('exec-list');
-    execList.removeItem(0);
-    hasEmptyItem = false;
+function hideEmptyLabel() {
+    document.getElementById('empty-label').classList.add('cordova-hidden');
+    document.getElementById('exec-list').classList.remove('cordova-hidden');
```
albertinad: Maybe it is better to save the `empty-label` and `exec-list` elements into variables to re-use, instead of querying the DOM always. We don't need to access these elements properties just one time for initialization only, but also given certain conditions.

albertinad:
This looks good to me! just minor comments.

albertinad:
Looks good to me!

Issue #63, opened 2016-06-09
[confusing ui] I'm not sure what's the Go 16x button is for.


digeff:
First time I used it I assume it might be the zoom level of the map.
It doesn't seem to be changing that, so now I'm guessing it's something related to the Navigator Simulator (probably the reproduction speed), but it'd be nice if the intent was clearer...

dlebu:
@guillaumejenkins @albertinad I agree this can be misleading. Would it make sense to put the `Choose File` button in the same row as the `Go` and `16x` button?

albertinad:
Yes! it would be much more clear.

PR #82, opened 2016-06-13
Fixing a couple of bugs

dlebu:
#63 [confusing ui] I'm not sure what's the Go 16x button is for. 

- I moved the `Choose file` button on the same row as `Go` and `16x`

https://github.com/Microsoft/vscode-cordova/issues/135 "How do I create a GPX file?" link is not working for me

- I removed the GPX file link and the send feedback one, since they are not functional inside vscode


albertinad:
LGTM!

PR #84, opened 2016-06-15
[review-discussion] Server modules re-design 

albertinad:
@dlebu @guillaumejenkins @busykai The intention of this PR is to open the discussion about a re-design that we can make to the server modules migrating to a more object oriented design, helping us to solve some problems:
- Today the state remains at module level, so for cleaning up the state in order to re-launch a new simulation, we need to go module by module and export an API or do something to reset the state, and it is not safe
- When stopping the simulation and launching a new one, no matter if it is for the same project or a new one with different/same configuration, some async tasks like prepare or creating the simulation files may not finish yet, and we have already stopped, so when it finish, there's a new context, and we can have multiple failures
- We can't create multiple simulation instances with an isolated context that is related to each individual simulation

There might be many ways to address these issues, but I started with this proposal to discuss, in which we have a couple of objects that will interact during the simulation, as described by this diagram:

server-design

Note: I don't add all the return types and parameters to the functions, to keep it simple.

I need to merge new changes in master, and manually migrate them, but the point is to present this proposal and open the discussion :) then I will start taking changes from master in order to keep it up to date.

Thanks!


src/server/config.js
```diff
@@ -1,51 +1,49 @@
...
     { name: 'forcePrepare' },
     { name: 'liveReload' },
     { name: 'platform' },
-    { name: 'platformRoot' },
-    { name: 'projectRoot', single: true },
-    { name: 'server', optional: true },
+    // NOTE: platformRoot and projectRoot belongs to the Project instance
+    // { name: 'platformRoot' },
+    // { name: 'projectRoot', single: true },
+    // { name: 'server', optional: true },
```
albertinad: I leave these lines commented just for information about the changes done, but it should be removed.

src/server/config.js
```diff
@@ -1,51 +1,49 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 
-var fs = require('fs'),
-    path = require('path');
-
-var config = {};
+/**
+ * @constructor
```
albertinad: Note that some constructor functions lacks of documentation, I will write it.

dlebu:
I took a look at the diagram for the proposed design changes, but not at the code yet. Here are my questions:
	• Simulator has an `appUrl` property - should we start thinking about accommodating multiple `APP` per `SIM`?
	• I see a circular dependency between `SimulationServer` and `Simulator`. Why is this needed?
	• Same question with `SocketServer` and `Simulator` (`Simulator` -> `SimulationServer` -> `SocketServer` -> `Simulator`). I'm trying to understand whether we can consolidate these entities into bigger ones or change the dependency chain.
	• Some of the `Project` properties are found in `Configuration` as well. Would it make sense for the `Project` to depend on `Configuration` directly, without re-declaring the properties?

src/server/live-reload/watcher.js
```diff
@@ -1,27 +1,37 @@
...
 }
 
 Watcher.prototype.startWatching = function () {
-    this.wwwWatcher = fs.watch(path.join(config.projectRoot, WWW_ROOT), { recursive: true }, handleWwwWatcherEvent.bind(this));
+    var watchPath = path.join(this.projectRoot, WWW_ROOT);
+    this.wwwWatcher = fs.watch(watchPath, { recursive: true }, handleWwwWatcherEvent.bind(this));
 
     if (this.mergesOverrideExists) {
-        this.mergesWatcher = fs.watch(path.join(config.projectRoot, MERGES_ROOT, config.platform), { recursive: true }, handleMergesWatcherEvent.bind(this));
+        watchPath = path.join(this.projectRoot, MERGES_ROOT, this.platform);
```
guillaumejenkins: No need to reconstruct the path here, we can use `this.mergesOverridePath` directly

src/server/project.js
```diff
@@ -0,0 +1,291 @@
...
+Project.prototype.prepare = function() {
+    if (!this._preparePromise) {
+        var d = Q.defer();
+        var currentProjectState;
+
+        this._preparePromise = d.promise;
+        this._lastPlatform = this.platform;
+
+        this._getProjectState().then(function (currentState) {
+            currentProjectState = currentState || this._getProjectState();
```
guillaumejenkins: The `|| this._getProjectState()` here is actually a bug in my previous code, as that method returns a promise and not a value. It can be safely removed, as `currentState` will always contain the result of calling `this.getProjectState()`, so it should never be undefined.

So, this line will simply become `currentProjectState = currentState;`

src/server/project.js
```diff
@@ -0,0 +1,291 @@
...
+                    return true;
+                });
+            }
+
+            return Q(false);
+        }.bind(this)).then(function (didPrepare) {
+            var previousState = this._previousPrepareStates[this.platform];
+
+            if (didPrepare || pluginUtil.shouldInitPlugins(currentProjectState, previousState, this.platform)) {
+                this._previousPrepareStates[this.platform] = currentProjectState || this._getProjectState();
```
guillaumejenkins: Same thing here, we should remove the `|| this._getProjectState()`

src/server/project.js
```diff
@@ -0,0 +1,291 @@
...
+    if (!this._preparePromise) {
+        var d = Q.defer();
+        var currentProjectState;
+
+        this._preparePromise = d.promise;
+        this._lastPlatform = this.platform;
+
+        this._getProjectState().then(function (currentState) {
+            currentProjectState = currentState || this._getProjectState();
+            var previousState = this._previousPrepareStates[this.platform];
```
guillaumejenkins: This variable could be refactored out of the promise chain so that it can be reused further down instead of defining it again. This is not a huge deal though, just something that crossed my mind :)

src/server/project.js
```diff
@@ -0,0 +1,291 @@
...
+    prepareUtil = require('./utils/prepare'),
+    utils = require('./utils/jsUtils');
+
+/**
+ * @constructor
+ */
+function Project(simulator, platform, debugHostHandlers) {
+    this._simulator = simulator;
+    this.platform = platform;
+    this.debugHostHandlers = debugHostHandlers;
```
guillaumejenkins: I notice that `this.debugHostHandlers` seems static. The intent is that if a debug-host disconnects from the server, this should be updated to properly reflect that there is no longer any debug-host, so on the next prepare / refresh, the debug-specific panels will not be added to sim-host. Does the current `Project` class handle debug-hosts disconnecting?

albertinad:
Thanks for your comments! I'm in the process of updating this branch with latest master, It will take me a while, once it is ready, I will address all of your comments :)

src/server/live-reload/watcher.js
albertinad: That's right! thanks!

src/server/project.js
albertinad: I've updated this branch with the latest changes on master, and I've integrated those fixes in the new model. Thanks!

src/server/project.js
albertinad: You're right, I did a minor modification here. Thanks!

src/server/project.js
albertinad: `debugHostHandlers` was duplicated, so I removed from the `Project` model in favor of keeping it as part of the `Simulator` model, since it looks like more to a configuration related to the simulation itself rather than the project.
Thanks!

albertinad:
@dlebu here are some responses to your questions:

> Simulator has an appUrl property - should we start thinking about accommodating multiple APP per SIM?
I'm not so sure if I understand correctly this question, but here goes: I don't see what is the relationship between the appUrl and multiple app-host per sim-host. The appUrl could be the same for multiple app-host connections. I think that we should focus on the new changes for this PR, prepare the infrastructure for supporting multiple app host that is almost ready, and then focus on that specific change.

> I see a circular dependency between SimulationServer and Simulator. Why is this needed?
The main reason of having these kind of bi-directional association is to enable some instances to access to some properties and functionality that are related to the overall simulation, like `config`, `project` and APIs that enables to check the status of the simulation: `isActive` etc. We can definitely improve this.

> Same question with SocketServer and Simulator (Simulator -> SimulationServer -> SocketServer -> Simulator). I'm trying to understand whether we can consolidate these entities into bigger ones or change the dependency chain.
Agree, let me review the dependency chain and see how can I improve it. I'm open to suggestions :)

> Some of the Project properties are found in Configuration as well. Would it make sense for the Project to depend on Configuration directly, without re-declaring the properties?
Some of the properties that used to be defined in the `config` module are now defined as properties of the `Project` model and some of them are duplicated, but that is my mistake. My intention was to keep everything specific to the project, in the `Project` model, but I forgot to remove it from the `Configuration` object. I have my doubts about the `Configuration` object. I keep it only for an easy mapping with the current implementation, and the new changes, but I think that we can directly define each property as part of the `Simulator` model, directly, instead of the `Configuration` object. What do you think about it?

dlebu:
@albertinad Thanks for the clarifications.

>      I'm not so sure if I understand correctly this question, but here goes: I don't see what is the relationship between the appUrl and multiple app-host per sim-host. The appUrl could be the same for multiple app-host connections. I think that we should focus on the new changes for this PR, prepare the infrastructure for supporting multiple app host that is almost ready, and then focus on that specific change.

You are right, for some reason I was thinking about multiple, (and different) apps and one SIM.
> I think that we can directly define each property as part of the Simulator model, directly, instead of the Configuration object. What do you think about it?

I think it would be good if it is possible. If `Configuration` is only used by simulator, we should definitely do it.

albertinad:
@dlebu since `Configuration` is used in other instances too, and exposed as a public property, then I think we should keep it as a separate entity, as you said, to get a better understanding of the configuration options for the simulation.

albertinad:
@dlebu @guillaumejenkins what do you think if we try to include these changes as part of the milestone 0.2.1? I believe that is important since it fixes several problems with the context of the simulation when it is stopped/started, and it will help us to start thinking on testing at unit and integration level.
What do you think? :)
I need to merge latest master into this branch, and a couple of more improvements.

src/server/live-reload/live-reload-server.js
```diff
@@ -1,32 +1,58 @@
...
 var Q = require('q');
-var telemetry = require('../telemetry-helper');
 var Watcher = require('./watcher').Watcher;
 
-var socket;
-var watcher;
+/**
+ * @constructor
+ */
+function LiveReload(project, telemetry, forcePrepare) {
```
dlebu: Would `LiveReloadServer` be a better name?

src/server/project.js
```diff
@@ -0,0 +1,346 @@
...
+    dirs = require('./dirs'),
+    pluginUtil = require('./utils/plugins'),
+    prepareUtil = require('./utils/prepare'),
+    utils = require('./utils/jsUtils');
+
+/**
+ * @constructor
+ */
+function Project(simulator, platform) {
+    this._simulator = simulator;
```
dlebu: The `_simulator` is used in this file only for its `telemetry` and `config` properties. Instead of passing the `_simulator` object here, can we pass only what is needed - its `telemetry` and `config` properties?

src/server/telemetry.js
```diff
@@ -13,16 +12,30 @@ var PLUGIN_PROPS_TO_ANONYMIZE = [
 ];
 
 /**
- * Handles telemetry sent by clients. Anonymizes properties if necessary, and sends the event to the external telemetr module if one was provided.
- * 
+ * @param {object=} opts
+ * @constructor
+ */
+function Telemetry(simulator, opts) {
+    this._simulator = simulator;
```
dlebu: Same here. We only need the `project`, not the parent `simulator` object.

src/server/simulator.js
```diff
@@ -0,0 +1,187 @@
...
+            // Get dynamically so simHostOptions is initialized
+            return simHostOptions.simHostRoot;
+        }
+    });
+
+    var platform = opts.platform || 'browser';
+
+    this._telemetry = new Telemetry(this, this._config.telemetry);
+    this._project = new Project(this, platform);
+    this._server = new SimulationServer(this);
```
dlebu: `Simulator`, `SimulationServer` and `SocketServer` are very tightly coupled. I wonder if we could do better. If only JS supported interfaces. 🎱

src/server/live-reload/live-reload-server.js
albertinad: I was thinking that maybe we can rename the file to `live-reload.js`, since this is not handling any server.

src/server/simulator.js
albertinad: I have a proposal for decoupling this, instead of passing through parameters `this`, and generating a circular dependency, I can create a kind of proxy object that contains the API that we want to expose for the instances, and that is a subset of all of the APIs. Doing this will help us too to testing the objects, without the need of a full mock of the main instance. I will give a try and then share to you for your comments :)

dlebu: Sounds good. Thanks!

src/server/project.js
albertinad: I have change this a little, the name remain the same, maybe we can have a better name for that property, but what is receives through parameters it is a kind of proxy object to the simulator APIs that all internal instances needs until today: telemetry and config. So we break circular dependency here.

src/server/telemetry.js
albertinad: This change is pending, I'm improving how telemetry module is created and initialized so I can pass only the project. 
At the point in which this object is created, the project instance is not ready yet, and in order to create a project instance, we need the telemetry instance ready.

src/server/simulator.js
```diff
@@ -0,0 +1,194 @@
...
+    var platform = opts.platform || 'browser';
+
+    var telemetry = new Telemetry(this, this._config.telemetry);
+
+    // create an intermediate object that expose only the
+    // required public API for the simulation objects
+    var simulatorProxy = {
+        config: this.config,
+        telemetry: telemetry
+    };
```
albertinad: @dlebu with this object, we break the circular dependency in most cases

src/server/project.js
```diff
@@ -0,0 +1,348 @@
...
+
+Project.prototype.getRouter = function () {
+    this._router = this._router || cordovaServe.Router();
+    return this._router;
+};
+
+Project.prototype.hasBuiltInPluginTelemetry = function (plugin) {
+    var builtIn = this.pluginsTelemetry.simulatedBuiltIn;
+
+    return (builtIn.indexOf(plugin) !== -1);
```
albertinad: @guillaumejenkins I've added this new function that checks if the given plugin has telemetry registered (built-in plugin). This logic was implemented in the telemetry module, when checking if the telemetry should be anonymized

src/server/server.js
```diff
@@ -6,43 +6,168 @@ var fs = require('fs'),
...
+ * Web Socket server. It enables to start and stop the server to listening
+ * for new connections.
+ * @param {object} simulator
+ * @constructor
+ */
+function SimulationServer(simulator, project, hostRoot) {
+    this._simulator = simulator;
+    this._project = project;
+    this._hostRoot = hostRoot;
+    this._cordovaServer = cordovaServe();
```
albertinad: I made a small change here, renaming `this._server` to `this._cordovaServer`. It is more representative of the type of instance we are holding.

src/server/project.js
dlebu: Sounds good, thanks!

dlebu: Can we rename this parameter and property to `_simulatorProxy` such that there is no confusion? Another option could be to pass the `telemetry` and `config` directly.

albertinad: That sounds good, I will change it to `_simulatorProxy` :)
Doing what you suggest is a good option, but there are some points that I consider when using this proxy object that acts as a kind of interface:
- telemetry and config are widely used across all of the other instances, and big candidates to still be widely used if we are new models/objects to the server
- if I pass through parameter both of them, on each instance that uses it, I'm duplicating information in all over the places, when actually the config and telemetry are not likely to be properties of each single object, but to the simulator itself, who coordinates the interaction between all of the components
- The simulatorProxy acts a middle instance that provides common functionality across all of the objects, and I guess that this approach is more flexible and will provide us a clean way to even add more functionality. Today we need what we have right now, but we can easily expose status APIs, so other instances checks the status before performing critical actions that modifies or generates files as a result of asynchronous operations, for example (thinking out loud)

dlebu: Cool. With telemetry and config being used together a lot, we might want to make telemetry a new config property. We don't need to do it now, but if it makes sense, we could do it in the future.

src/server/server.js
```diff
@@ -6,43 +6,176 @@ var fs = require('fs'),
...
 
-var pluginSimulationFiles = require('./plugin-files');
+/**
+ * The simulation server that encapsulates the HTTP server instance and
+ * Web Socket server. It enables to start and stop the server to listening
+ * for new connections.
+ * @param {object} simulator
+ * @constructor
+ */
+function SimulationServer(simulator, project, hostRoot) {
```
dlebu: Same here. Please rename to `simulatorProxy`

dlebu:
This looks good to me @albertinad ! Thanks for making this change! I had minor comments, please consider me signed off. :shipit:

src/server/server.js
albertinad: Sure! done :)

albertinad:
Thank you @guillaumejenkins and @dlebu! I've addressed all of your comments :)

albertinad:
I've pushed the latest changes to this PR:
- documentation for each new constructor function added
- improvement in how the state is handled when starting/stopping the simulation
- error handling for some `server.close` calls that might throw an exception

Thanks!

src/server/project.js
TimBarham: I'm thinking 'simulatorProxy` is the wrong terminology here. That would imply this is some sort of intermediary between the simulator and other things. But really it is more of a "context" type object, right? How about `simulatorContext` or `simulationContext`?

albertinad: `simulatorProxy` is an object that provides the required interface for getting the state of the simulation. So yes, it is more like a context object, if by context we can also mean to exposing functions to get more information and also functions that could change the behaviour of the `simulator` instance. I don't have a clear example for the latest. But if `simulatorContext` give us a better understanding of the meaning of this object, then let's change it!

src/server/config.js
```diff
@@ -1,48 +1,47 @@
...
-var config = {};
+/**
+ * The Configuration model manages the configuration properties for the simulation.
+ * These properties are mapped with the options provided for the simulation, and
+ * are cross to any other instance involved in the simulation.
+ * @constructor
+ */
+function Configuration() {
+    this._config = {};
+}
```
busykai: Maybe it would be nicer to not inherit from `Object`, but not have a prototype at all:

```javascript
Configuration.prototype = Object.create(null);
Configuration.prototype.constructor = Configuration;
```

albertinad: This is nice! I'm thinking that, it would be nicer too if we don't have `this._config` property, we can attach the properties directly to the instance.

src/server/config.js
```diff
@@ -1,48 +1,47 @@
...
-            setValue(prop.name, value, prop.single);
+            setValue(this, prop.name, value, prop.single);
         }
     });
 });
 
-function setValue(prop, value, single) {
-    if (single && config.hasOwnProperty(prop)) {
+function setValue(instance, prop, value, single) {
+    if (single && instance._config.hasOwnProperty(prop)) {
```
busykai: If the property has been defined previously, then re-configuring it (e.g. specifying another setter) will throw `TypeError` before you get a chance to execute this code. I would instead recommend a more "javascriptish" way of applying `setValue()` function to the context of the call (and not passing `this` explicitly):

```javascript
Object.defineProperty(Configuration.prototype, prop.name, {
    get: function () {
        return getValue.apply(this, [prop.name, prop.optional]);
    },
    set: function (value) {
        return setValue.apply(this, [prop.name, value, prop.single]);
    }
}
```

src/server/config.js
```diff
@@ -1,48 +1,47 @@
...
-    { name: 'projectRoot', single: true },
-    { name: 'server', optional: true },
     { name: 'simHostOptions' },
     { name: 'simulationFilePath' },
     { name: 'telemetry', optional: true },
     { name: 'touchEvents', optional: true },
     { name: 'xhrProxy', optional: true }
 ].forEach(function (prop) {
-    Object.defineProperty(module.exports, prop.name, {
+    Object.defineProperty(Configuration.prototype, prop.name, {
```
busykai: `Object.definedProperty()` will cause `TypeError` in this form (re-defining setter and getter). This should be wrapped into a `try/catch`, expecting a `TypeError`.

src/server/config.js
busykai: Yes, `_config` does not seem to be necessary after you made configuration an object.

src/server/config.js
```diff
@@ -1,48 +1,47 @@
...
         throw new Error('Can\'t reinitialize ' + prop);
     }
-    config[prop] = value;
+    instance._config[prop] = value;
 }
 
-function getValue(prop, optional) {
-    if (!config.hasOwnProperty(prop) && !optional) {
+function getValue(instance, prop, optional) {
+    if (!instance._config.hasOwnProperty(prop) && !optional) {
```
busykai: Just a word of caution: if you remove `_config` and inherit from `Object.create(null)`, then `hasOwnProperty` will not work. You can, though, instead rely on the fact that the properties you define on `Configuration.prototype` itself are non-configurable (and specify it explicitly).

src/server/config.js
busykai: Same inheritance chain could be applied to other modules converted to objects (like `LiveReload` and such).

src/server/server.js
```diff
@@ -77,21 +223,30 @@ function streamAppHostHtml(request, response) {
...
+ */
+SimulationServer.prototype._streamSimHostHtml = function (request, response) {
     // If we haven't ever prepared, do so before we try to generate sim-host, so we know our list of plugins is up-to-date.
     // Then create sim-host.js (if it is out-of-date) so it is ready when it is requested.
-    prepare.prepare().then(function () {
-        return simFiles.createSimHostJsFile();
-    }).then(function () {
+    var project = this._project;
+
+    project.prepare().then(function () {
```
busykai: it would be nice if the call to `.then()` and `.done()` were on the next line (here and below). I believe in other places it was indented also, like this:

```javascript
project.prepare()
     .then(function () {
     }.bind(this))
...
     .done()
```

busykai:
I had just a few insignificant (non-functional) comments, if this is ready to merge, they are not blocking and could be addressed later.

src/server/server.js
albertinad: I prefer the indentation style you propose, it is better for reading it. In many places it was intended like the code in this review. @dlebu @guillaumejenkins @TimBarham @MSLaguana what do you think if we go with this indentation style that @busykai is proposing for promises chain? In my opinion, it is better for readability and easily understanding/following the promises chain.

TimBarham: Yep, I agree with @busykai - each `.then()` / `.done()` etc on a new line.

TimBarham: (or we switch to TypeScript and use `async`/`await` and avoid the problem altogether :smile:)

albertinad:
Thanks all of you! I've addressed a couple of the latest suggestions and comments, if there's no more comments in a couple of hours, I will merge it to master.

albertinad:
Merged! thanks!

src/server/project.js
TimBarham: This change didn't seem to make it in?

albertinad: Not yet, I will do it as part of next commits :) what do you think about this change @dlebu @guillaumejenkins @busykai @MSLaguana ?

PR #85, opened 2016-06-15
Remember collapse state of panels across sessions

### LINKED ISSUE #76: Remember collapsed / expanded state of panels across sessions
We should consider saving the collapsed / expanded state of the panels across simulate sessions. For users with a lot of plugins in their project, this could be very useful.

guillaumejenkins:
Fixes #76

### DIFF
--- a/src/sim-host/protocol/socket.js
+++ b/src/sim-host/protocol/socket.js
@@ -2,8 +2,9 @@
 
 /*global io: false */
 
-var telemetry = require('telemetry-helper'),
-    simStatus = require('sim-status');
+var Q = require('q'),
+    simStatus = require('sim-status'),
+    telemetry = require('telemetry-helper');
 
 var registerOnInitialize = false;
 var socket;
@@ -36,6 +37,8 @@ Object.defineProperty(module.exports, 'socket', {
 });
 
 module.exports.initialize = function (pluginHandlers, services) {
+    var deferred = Q.defer();
+
     serviceToPluginMap = services;
     socket = io();
 
@@ -45,6 +48,18 @@ module.exports.initialize = function (pluginHandlers, services) {
 
     socket.on('refresh', function () {
         document.location.reload(true);
+    });
+
+    socket.on('connect', function () {
+        deferred.resolve();
+    });
+
+    socket.on('connect_error', function (err) {
+        deferred.reject(err);
+    });
+
+    socket.on('connect_timeout', function (err) {
+        deferred.reject(err);
     });
 
     socket.on('app-plugin-list', function () {
@@ -102,6 +117,8 @@ module.exports.initialize = function (pluginHandlers, services) {
     if (registerOnInitialize) {
         registerSimHost();
     }
+
+    return deferred.promise;
 };
 
 module.exports.notifyPluginsReady = function () {
--- a/src/sim-host/ui/sim-host.js
+++ b/src/sim-host/ui/sim-host.js
@@ -1,28 +1,61 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 
-var db = require('db'),
+var customElements = require('./custom-elements'),
+    db = require('db'),
+    dialog = require('dialog'),
     Messages = require('messages'),
-    customElements = require('./custom-elements'),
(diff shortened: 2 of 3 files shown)


src/sim-host/protocol/socket.js
```diff
@@ -80,10 +83,21 @@ module.exports.initialize = function (pluginHandlers, services) {
...
     });
+    socket.on('connect', function () {
+        deferred.resolve();
+    });
+    socket.on('connect_error', function (err) {
+        deferred.reject(err);
+    });
+    socket.on('connect_timeout', function (err) {
+        deferred.reject(err);
+    });
```
albertinad: @busykai  will these changes impact something related to the new protocol?

guillaumejenkins: This is just to make the function return a promise so it can be correctly chained with other promises. It shouldn't impact anything, as consumers can still ignore the returned promise if they want this call to happen purely asynchronously.

src/sim-host/ui/sim-host.js
```diff
@@ -1,29 +1,68 @@
...
+    }
+
+    db.saveObject(COLLAPSED_PANELS_KEY, collapsedPanels);
+}
+
+function getCollapsedPanels() {
+    var collapsedPanels;
+
+    try {
+        collapsedPanels = db.retrieveObject(COLLAPSED_PANELS_KEY);
```
dlebu: Can we query for the entry before retrieving, instead of relying on the error mechanism?

guillaumejenkins: The error comes from the `retrieveObject` expecting an object, not from the value not being present... if the value is not in the local storage, `undefined` is simply returned. I'm not sure this check is a whole lot useful... it was just in case the storage did not contain an array, for extra protection

dlebu: Sounds good.

dlebu:
Looks good to me.

src/sim-host/ui/sim-host.js
```diff
@@ -1,29 +1,68 @@
...
+    } catch (err) {
+        // Do nothing, default empty array will be used.
+    }
+
+    if (!collapsedPanels || !Array.isArray(collapsedPanels)) {
```
albertinad: the first validation is not needed, since you're checking using Array.isArray. If null or undefined is passed to isArray, it will return false.

albertinad:
Looks good to me, just a minor comment.

src/sim-host/ui/sim-host.js
guillaumejenkins: `db.retrieveObject` throws an error if the value is defined, but is not an Object, so the `catch` is needed, and then the value might be a pure Object and not an array, so the `isArray` is needed as well. However, we don't have this kind of protection anywhere else, so I don't know if it makes sense to have them here...

albertinad: So `collapsedPanels` is an Array for us, then we just need to do this:
```
if (!Array.isArray(collapsedPanels)) {
  collapsedPanels = [];
}
```

If the db.retrieveObject returns an Object, not an array, then it will evaluate to true in the condition that we are using (above) and an empty array will be used.

guillaumejenkins: Sure I can do that, we'll still get an uncaught Error if the stored value is not an Object or an Array, but I don't think that should happen under normal circumstances

albertinad: Sounds good! 
I was taking a look at the implementation of [retrieveObject](https://github.com/Microsoft/cordova-simulate/blob/master/src/modules/db.js#L122), and the exception could be produced by JSON.parse, so maybe we can handle the exception in there, and return some default value, null or undefined. To avoid delegating the responsibility of handling the error of JSON.parse to the consumers of this API. But this is not needed to be done as part of this PR, we can think about it fix it later, just in case we miss something.

guillaumejenkins: Yes it makes a lot of sense that this should be handled by `db`. And you are right, the error I was getting was `Unexpected token '3'` when I had a number as the saved value, so this is definitely coming from the `JSON.parse`

PR #91, opened 2016-06-21
Camera support for DATA_URL destination type.

### LINKED ISSUE #34: simulate camera plugin with DATA_URL is not working
Trying to run simulate on this sample https://blog.ionic.io/10-minutes-with-ionic-2-using-the-camera-with-ionic-native/

I got the error 

Failed to load resource: net::ERR_INVALID_URL (data:image/png;base64,blob:http%3A//localhost%3A8000/72ab5a1f-f816-4a3e-bdd9-57b9395a8b04)

dlebu:
Fixes #34

### DIFF
--- a/src/plugins/cordova-plugin-camera/app-host-handlers.js
+++ b/src/plugins/cordova-plugin-camera/app-host-handlers.js
@@ -5,9 +5,14 @@ module.exports = function (messages) {
         Camera: {
             takePicture: function (success, fail, args) {
                 messages.call('takePicture', args).then(function (result) {
-                    // 'result' should be {data: <ArrayBuffer>, type: <mimeType>}, from which we'll create a blob
-                    var blob = new Blob([result.data], {type: result.type});
-                    success(URL.createObjectURL(blob));
+                    if (args && args[1] === 0) {
+                        /* Destination type is DATA_URL */
+                        success(result);
+                    } else {
+                        // 'result' should be {data: <ArrayBuffer>, type: <mimeType>}, from which we'll create a blob
+                        var blob = new Blob([result.data], { type: result.type });
+                        success(URL.createObjectURL(blob));
+                    }
                 }, function (error) {
                     fail(error);
                 });
--- a/src/plugins/cordova-plugin-camera/sim-host.js
+++ b/src/plugins/cordova-plugin-camera/sim-host.js
@@ -22,7 +22,7 @@ module.exports = function (messages) {
                     // handler with one that uses the appropriate value of 'callback' from the current closure.
                     document.getElementById('camera-dialog-use-image').onclick = function () {
                         dialog.hideDialog('camera-choose-image');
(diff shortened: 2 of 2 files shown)


src/plugins/cordova-plugin-camera/app-host-handlers.js
```diff
@@ -5,9 +5,14 @@ module.exports = function (messages) {
         Camera: {
             takePicture: function (success, fail, args) {
                 messages.call('takePicture', args).then(function (result) {
-                    // 'result' should be {data: <ArrayBuffer>, type: <mimeType>}, from which we'll create a blob
-                    var blob = new Blob([result.data], {type: result.type});
-                    success(URL.createObjectURL(blob));
+                    if (args && args[1] === 0) {
```
guillaumejenkins: While we're here, should we support the 3rd destination type as well?

### Camera.DestinationType : <code>enum</code>
**Kind**: static enum property of <code>[Camera](#module_Camera)</code>  
**Properties**

| Name | Type | Default | Description |
| --- | --- | --- | --- |
| DATA_URL | <code>number</code> | <code>0</code> | Return base64 encoded string. DATA_URL can be very memory intensive and cause app crashes or out of memory errors. Use FILE_URI or NATIVE_URI if possible |
| FILE_URI | <code>number</code> | <code>1</code> | Return file uri (content://media/external/images/media/2 for Android) |
| NATIVE_URI | <code>number</code> | <code>2</code> | Return native uri (eg. asset-library://... for iOS) |

dlebu: You are right, there is no work to be done here for the NATIVE_URI destination type. For both FILE_URI and NATIVE_URI we will return a HTTP URL which will then be displayed in APP_HOST accordingly. The only exception is the DATA_URL for which we need to return the base64 encoded image as a string.

guillaumejenkins: Great!

andreszs:
@dlebu @guillaumejenkins  According to the [cordova-plugin-camera readme](https://github.com/apache/cordova-plugin-camera#browser-quirks), the **browser** version can only return base64 encoded images. However, I'm getting this now:
```
blob:http://localhost:8000/259e6e9e-bb2f-47d0-939d-51461bd7d47c
```

As a bonus, `targetWidth` and `targetHeight` parameters are ignored. Is the documentation wrong, or what's the problem here? I'm using camera plugin version 2.4.2-dev (master), CLI 6.5.0, and Chrome. Thanks

dlebu:
@andreszs Can you file a bug? Someone will take a look at it. Me and @guillaumejenkins are not actively working on this project any more. @TimBarham Any ideas?

andreszs:
@dlebu Hi, the bug has been filed under [CB-13316](https://issues.apache.org/jira/browse/CB-13316).

PR #93, opened 2016-06-21
Code of conduct

dlebu:


busykai:
Looks good to me. I need to read it.

I think this needs to be squashed so that `.vscode/` part does not appear in the history. Should `.vscode/` be added to `.gitignore`?

albertinad:
Looks good to me! I need to read it too.

dlebu:
@busykai I squashed things together. Thanks!

PR #94, opened 2016-06-21
Fix #90.

### LINKED ISSUE #90: SIM_HOST plugin code may send messages ahead of time
In some cases, plugin simulation code on SIM_HOST may send messages ahead of time (before the APP_HOST is able to respond). Currently, there's one specific case: `cordova-plugin-device` tries to retrieve `cordova.version` ahead of time. With the merge of #21, it becomes consistently reproducible (in Device widget cordova version information is missing). Since the call never returns, the `.fail()` handler on such calls is not called either.

busykai:
Provide an API for the plugins to synchronize calls to the APP_HOST.

We had a few options to address issue #90. One of them was to simply delay premature messages to APP_HOST until it's ready on the server and the other embed plugin initialization into the lifecycle of the entire simulation.

This is an alternative approach which provides an API to the plugins to sync on for the messages which should be sent to the APP_HOST early in the initialization. I do think that unsolicited (by APP_HOST) messages and calls from SIM_HOST to APP_HOST should be a rare exception, but if they absolutely have to be sent, there's an API which allows them to do that in a coordinated manner. It does not impose any change on the plugins which do not need, just the ones that do need to use it.

Even the use case we have right now (`cordova-platform-device` plugin), should probably not be the case. Instead, `cordova.version` (and `cordova` in general) information should be returned to SIM_HOST with `app-plugin-list` message during the bootstrap.

Please let me know what you think.

@dlebu, @guillaumejenkins, @albertinad

### DIFF
--- a/src/modules/sim-host/sim-status.js
+++ b/src/modules/sim-host/sim-status.js
@@ -1,0 +1,31 @@
+// Copyright 2016 Intel Corporation. All rights reserved.
+
+var isAppHostReady = false,
+    appHostReadyHandlers = [];
+
+function whenAppHostReady(handler) {
+    var idx;
+    if (typeof handler !== 'function') {
+        return;
+    }
+    idx = appHostReadyHandlers.push({'handler': handler, 'fired': false});
+    if (isAppHostReady) {
+        handler();
+        appHostReadyHandlers[idx-1].fired = true;
+    }
+}
+
+function fireAppHostReady() {
+    isAppHostReady = true;
+    appHostReadyHandlers.forEach(function (element) {
+        if (!element.fired) {
+            element.handler();
+            element.fired = true;
+        }
+    });
+}
+
+/* Internal use function */
+module.exports._fireAppHostReady    = fireAppHostReady;
+/* Public API */
+module.exports.whenAppHostReady     = whenAppHostReady;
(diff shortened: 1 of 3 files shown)


dlebu:
This looks good to me.

src/modules/sim-host/sim-status.js
```diff
@@ -0,0 +1,31 @@
+// Copyright 2016 Intel Corporation. All rights reserved.
```
albertinad: Should we add Microsoft license header too?

albertinad:
Looks good to me!

src/modules/sim-host/sim-status.js
busykai: yes, sorry. @dlebu are you ok with both copyrights?

dlebu: @busykai I will ask our legal team and get back to you.

busykai: @dlebu, should we hold it for it? I'd rather changes this to the header the rest of the files have and then we'll figure out double copyright thing.

dlebu: @busykai Sounds good. I emailed our legal team but we can modify it later if needed. Feel free to merge.

busykai:
OK. Merging. Fixes #90.

PR #96, opened 2016-06-23
Add indent rule to linting.

busykai:
Fix existing findings.

### DIFF
--- a/src/plugins/cordova-plugin-vibration/app-host.js
+++ b/src/plugins/cordova-plugin-vibration/app-host.js
@@ -86,16 +86,14 @@ module.exports = function (messages) {
         body.css({position: 'fixed'});
 
         for (x = 1; x <= times; x++) {
-            body.animate({ left: -10 }, 5,
-                function () { checkForVibration(deferred, body);
-            })
-            .animate({ left: 0 }, 1)
-            .animate({ left: 10 }, 5)
-            .animate({ left: 0 }, 1)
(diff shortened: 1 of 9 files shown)


src/plugins/cordova-plugin-vibration/app-host.js
```diff
@@ -87,8 +87,9 @@ module.exports = function (messages) {
 
         for (x = 1; x <= times; x++) {
             body.animate({ left: -10 }, 5,
-                function () { checkForVibration(deferred, body);
-            })
+                function () {
```
guillaumejenkins:
```
body.animate({ left: -10 }, 5, function () {
    checkForVibration(deferred, body);
})
.animate(...)
...
```

maybe? It seems like everywhere else we've typically put `function () {` on the same line as the initial call

busykai: So these two variants make the linter happy:
1) invocations indented consistently
```javascript
body
    .animate({ left: -10 }, 5, function () {
        checkForVibration(deferred, body);
    })
    .animate(...)
```
or one-liner function:
2)
```javascript
body.animate(..., function () { checkForVibration(deferred, body); })
    .animate(...)
```
In this case I opted for the latter one (since the function body is short).

albertinad:
Looks good to me!

busykai:
Thanks! Merging.

Issue #86, opened 2016-06-15
Number entry in Compass panel is not correctly initialized when sim-host is launched


guillaumejenkins:
See following screenshot:

![image](https://cloud.githubusercontent.com/assets/14299377/16101406/2d957740-331a-11e6-95b0-5a9c55c1e2f4.png)

Changing the value on the 2D compass or in the geolocation panel makes the number correctly appear.

albertinad:
Fixed in #100

Issue #71, opened 2016-06-09
[weird behavior] In compass pressing the arrow down from 360 goes 359.99, 359, 358, etc...


digeff:
In compass pressing the arrow down from 360 goes 359.99, 359, 358, etc...
![image](https://cloud.githubusercontent.com/assets/14098140/15948828/edc684f4-2e57-11e6-9b20-a9b72504cccf.png)
![image](https://cloud.githubusercontent.com/assets/14098140/15948833/f5c3ecfa-2e57-11e6-906f-ab3af338bff7.png)
![image](https://cloud.githubusercontent.com/assets/14098140/15948840/fe337d1a-2e57-11e6-8f49-904bea04f914.png)

albertinad:
Fixed in #100

albertinad:
@dlebu I noticed that this issue is scheduled for vNext milestone right now, but I already fixed it with a couple of more fixes related to the compass.

dlebu:
@albertinad No problem. Thanks for the fixes!

PR #100, opened 2016-06-28
Several compass simulation fixes

albertinad:
Fix for #86 and #71. The proposed fix for the latest one is a combination of checking the maximum and minimum values that the compass heading can take, according the plugin spec, and also changing the input number step value to match the geolocation step: 0.5.


dlebu:
Looks good to me.

Issue #103, opened 2016-06-30
Geolocation: Changing long / lat values using arrows does not update the map


guillaumejenkins:
When changing longitude or latitude values by typing with the keyboard, the map gets properly updated to show the new coordinates. When using the arrows in the Number entry field to change the value, the map stays where it is.

![animation](https://cloud.githubusercontent.com/assets/14299377/16496886/1a4cd65c-3ea9-11e6-9504-1ceacb8d78a7.gif)

albertinad:
Fixed in #104

PR #104, opened 2016-06-30
Geolocation widget: fixing input number events

albertinad:
This PR fixes the issue #103, which I've introduced when I updated from an input HTMLElement to cordova input web component.

Thanks!

### DIFF
--- a/src/plugins/cordova-plugin-geolocation/sim-host-panels.html
+++ b/src/plugins/cordova-plugin-geolocation/sim-host-panels.html
@@ -1,6 +1,6 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
 
-<!-- Include OpenLayers here rather than require"ing it, since it doesn"t expose global variables when require"d -->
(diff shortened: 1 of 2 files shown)


src/plugins/cordova-plugin-geolocation/sim-host-panels.html
```diff
@@ -1,6 +1,6 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
 
-<!-- Include OpenLayers here rather than require'ing it, since it doesn't expose global variables when require'd -->
+<!-- Include OpenLayers here rather than require"ing it, since it doesn"t expose global variables when require"d -->
```
albertinad: Oops! wrong find and replace!

MSLaguana: Looks like some over-zealous find/replace happend here

MSLaguana:
Looks good to me.

PR #107, opened 2016-07-05
Windows API: Adding stub for Windows.UI.Notification

albertinad:
This PR is to add a missing stub for the windows platform API: Windows.UI.Notification.
I was trying some plugins different to the core plugins, and I found that this API is missing. I guess that there are more namespaces and functions that we are missing, we should have an strategy for that. Should we mock as much as we can? Or should we provide an easy way of configuring a default behaviour for this missing API?, for the developers that are using cordova-simulate, not providing more simulation for other plugins.

Thanks!

### DIFF
--- a/src/platforms/windows/app-host-clobbers.js
+++ b/src/platforms/windows/app-host-clobbers.js
@@ -8,7 +8,7 @@
 module.exports = {
     WinJS: {
         Application: {
-            addEventListener: function () { }
+            addEventListener: function () {}
         },
         Utilities: {
             // While simulating Windows platform, we don't currently provide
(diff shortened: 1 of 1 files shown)


MSLaguana:
Is this something which is normally provided by the webview? That is, does IE provide these interfaces?

If it is actually a platform difference, I'm not sure that we should hide it here. On the one hand on a real device these methods would be available, but on the other hand I don't think that we can correctly replicate platform quirks and it would be better for the user to run the simulation in a browser that more closely matches the target platform.

albertinad:
This is provided by the runtime. That's why we have this `platforms clobbers`. To provide some dummy implementation of the APIs that are not available in the browser. For windows based platforms, we have many missing APIs, because `Windows` and `WinJS` are not available in the regular browser.
I think that it is good if we can provide at least a minimum implementation, that at least give them the possibility to run their app, knowing the limitations. Sometimes, depending of the plugin, the exception that is thrown when trying to execute a function that is undefined, causes the plugin to not to load, and the app as a consequence will be in a broken state. I know we can't have a full control of these situations,  but at least we provide a way of running the app in a simulated environment, with limitations.

MSLaguana:
Ah, I see. In that case this change looks good to me.

dlebu:
LGTM!

PR #108, opened 2016-07-11
Validate input number to be always a number

albertinad:
According to the browser and version, sometimes a non-number is allowed in the input number field. This PR is for adding a validation to force any value of the `cordova-number-entry` to be a number. When a non-number value is added, or `""` the previous value is used to update the input value.
The `orientation` plugin doesn't use this component, so instead of that, I've added the common validation with the other validations that it has.

### DIFF
--- a/src/modules/utils.js
+++ b/src/modules/utils.js
@@ -283,6 +283,10 @@ self = module.exports = {
     },
 
     isNumber: function (value) {
+        if (typeof value === 'number') {
+            return true;
+        }
+
         return !!value.match(/-?(\d+|\d+\.\d+|\.\d+)([eE][-+]?\d+)?/);
     }
 };
--- a/src/modules/utils.js
+++ b/src/modules/utils.js
@@ -283,11 +283,9 @@ self = module.exports = {
     },
 
     isNumber: function (value) {
-        if (typeof value === 'number') {
-            return true;
-        }
+        var type = typeof value;
 
-        return !!value.match(/-?(\d+|\d+\.\d+|\.\d+)([eE][-+]?\d+)?/);
+        return (type === 'number' || type === 'string') && !isNaN(value - parseFloat(value));
     }
 };
 
--- a/src/modules/utils.js
+++ b/src/modules/utils.js
@@ -280,6 +280,10 @@ self = module.exports = {
             !url.href.match(/^https?:\/\/|^file:\/\//);
 
         return !!sameOrigin;
+    },
+
+    isNumber: function (value) {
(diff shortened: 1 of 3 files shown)


src/sim-host/ui/custom-elements.js
```diff
@@ -167,6 +168,7 @@ function initialize(changePanelVisibilityCallback) {
     registerCustomElement('cordova-number-entry', {
         value: {
             set: function (value) {
+                this._internalValue = value;
```
MSLaguana: Should this perform validation on `value` as well? Currently if the code using the `cordova-number-entry` is incorrect then it could assign a non-numeric value.

albertinad: Yes, I will update it. Thanks!

MSLaguana:
Looks good to me

src/modules/utils.js
```diff
@@ -280,6 +280,14 @@ self = module.exports = {
...
 
         return !!sameOrigin;
+    },
+
+    isNumber: function (value) {
+        if (typeof value === 'number') {
+            return true;
+        }
+
+        return !!value.match(/-?(\d+|\d+\.\d+|\.\d+)([eE][-+]?\d+)?/);
```
TimBarham: I think there are better ways to check for a number. For example, this will fail with European decimal notation (which uses a comma, not a period)? Some alternatives (that could replace the entire method):

    return !isNaN(parseFloat(value)) && isFinite(value);

Or the way jQuery does it (pre 3.0):

    return !Array.IsArray(value) && (value - parseFloat(value) + 1) >= 0;

jQuery 3.x:

    var type = typeof value;
    return (type === "number" || type === "string") && !isNaN(value - parseFloat(value));

MSLaguana: It depends on what we want to validate here. If we go with the `parseFloat` approach then `1.2hello` is a number equal to `1.2`. Similarly, with the current approach we forbid the european style `1,2`, but that would be incorrectly parsed as `1.0` if we allowed it.

TimBarham: None of the suggested methods will accept "1.2hello" as a number (that's why none of them *only* call `parseFloat()` :smile:). But yeah, you're right about the locale - none of these solutions handle that either (and it seems that is not a trivial task in JavaScript). Still, I think we're safer going with an approach like one of the above than using a regex.

albertinad: Agree that we can do better here, and support commas too. I took this regex from the validation that chrome does when a value is not a number, after a failure.  But supporting different notations, like european style, depends on i18n and l10n support that we don't have yet . We should add the support to all modules and plugins.

albertinad: I will update the validation to use the jQuery 3.x implementation.

PR #109, opened 2016-07-12
Fixing promise callback scope in SimulationServer.

albertinad:
This fixes a missing call to `bind` in a promise callback.


MSLaguana:
Looks good to me.

PR #110, opened 2016-07-12
Fix XHR proxy with non-standard ports

MSLaguana:
If a non-standard port is used, then the port will double-up at the moment, leading to a failure to contact `example.com:8080:8080`. Specifying `hostname` rather than `host` fixes this issue.


albertinad:
Looks good to me.

PR #113, opened 2016-07-13
Updating require usages in cordova-plugin-device-motion simulation.

albertinad:
The simulator server crashes when trying to generate the browserify context for a project that contains an old version of `cordova-plugin-device-motion` plugin, like version `0.2.11`. That old version has the old name of the plugin: `org.apache.cordova.device-motion`. I tried with different old plugins, and there was any problem, but with this one in particular, the server crashes with the following error:
```
/<folder>/cordova-simulate/node_modules/q/q.js:155
                throw e;
                      ^
Error: Cannot find module 'cordova-plugin-device-motion' from '/<folder>/cordova-simulate/src/plugins/cordova-plugin-device-motion'
    at /<folder>/cordova-simulate/node_modules/browserify/node_modules/resolve/lib/async.js:46:17
    at process (/<folder>/cordova-simulate/node_modules/browserify/node_modules/resolve/lib/async.js:173:43)
    at ondir (/<folder>/cordova-simulate/node_modules/browserify/node_modules/resolve/lib/async.js:188:17)
    at load (/<folder>/cordova-simulate/node_modules/browserify/node_modules/resolve/lib/async.js:69:43)
    at onex (/<folder>/cordova-simulate/node_modules/browserify/node_modules/resolve/lib/async.js:92:31)
    at /<folder>/cordova-simulate/node_modules/browserify/node_modules/resolve/lib/async.js:22:47
    at Object.oncomplete (fs.js:108:15)

```

The folders of for the plugin simulation exists, but in this particular case the module with the name of `cordova-plugin-device-motion`, that matches the plugin simulation folder and the plugin ID in the new style, doesn't exists in the browserify context. This error happens because the simulation plugin it is assuming that in browserify context, the name of the main entry point to the simulation module matches the folder and the ID in the new style, and try to [require it](https://github.com/Microsoft/cordova-simulate/blob/master/src/plugins/cordova-plugin-device-motion/sim-host-handlers.js#L3). But that seems to not always be true when the project has an old plugin. I guess that the name in browserify for this case is `org.apache.cordova.device-motion`.
As a proposal for fixing this issue, the simulation should not assume names of modules, so I changed the way the accelerometer axis values are used by `sim-host-handlers.js`: a new module that encapsulates the data model is added and used by `sim-host-handlers.js` and `sim-host.js`.

Besides that we don't official support simulation for different versions of plugins (built-in), I think that the server should not crash on this cases.

Thanks!

### DIFF
--- a/src/plugins/cordova-plugin-device-motion/device-motion-model.js
+++ b/src/plugins/cordova-plugin-device-motion/device-motion-model.js
@@ -1,0 +1,56 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+
+// report interval in milliseconds
+var ACCELEROMETER_REPORT_INTERVAL = 50;
+
+var _onChangeCallback,
+    _axis = {
+        x: 0,
+        y: 0,
+        z: 0
+    };
+
+function updateAxis(values) {
+    values = values || {};
+
+    var updatedValues = {};
+
+    if (values.x) {
+        _axis.x = parseFloat(values.x);
+        updatedValues.x = _axis.x;
+    }
+
+    if (values.y) {
+        _axis.y = parseFloat(values.y);
+        updatedValues.y = _axis.y;
+    }
+
+    if (values.z) {
+        _axis.z = parseFloat(values.z);
+        updatedValues.z = _axis.z;
+    }
+
+    // notify if any value has been updated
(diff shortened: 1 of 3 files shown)


MSLaguana:
Looks good to me.

albertinad:
Thanks! merging!

PR #115, opened 2016-07-14
Move prepare retry logic from LiveReload to execCordovaPrepare.

TimBarham:
Currently `LiveReload` will try a second time to prepare if the first time fails (usually because of errors trying to create www folder too quickly on Windows). This moves that logic into `execCordovaPrepare()` itself, so we gain the same benefit in other scenarios.

(I'm seeing this failure regularly on the first launch of simulations from VS Code)

### DIFF
--- a/src/server/utils/prepare.js
+++ b/src/server/utils/prepare.js
@@ -2,7 +2,8 @@
 
 var exec = require('child_process').exec,
     Q = require('q'),
-    log = require('./log');
+    log = require('./log'),
+    utils = require('./jsUtils');
 
 /**
  * @param {string} projectRoot
@@ -10,21 +11,27 @@ var exec = require('child_process').exec,
  * @return {Promise}
  */
 function execCordovaPrepare(projectRoot, platform) {
-    var deferred = Q.defer();
+    return utils.retryAsync(getExecCordovaPrepareImpl(projectRoot, platform));
+}
 
-    log.log('Preparing platform \'' + platform + '\'.');
+function getExecCordovaPrepareImpl(projectRoot, platform) {
+    return function () {
+        var deferred = Q.defer();
 
-    exec('cordova prepare ' + platform, {
-        cwd: projectRoot
-    }, function (err, stdout, stderr) {
-        if (err) {
-            deferred.reject(err || stderr);
-        }
+        log.log('Preparing platform \'' + platform + '\'.');
 
-        deferred.resolve();
-    });
+        exec('cordova prepare ' + platform, {
+            cwd: projectRoot
+        }, function (err) {
+            if (err) {
+                deferred.reject(err);
+            } else {
+                deferred.resolve();
+            }
+        });
 
-    return deferred.promise;
+        return deferred.promise;
+    };
 }
 
 module.exports.execCordovaPrepare = execCordovaPrepare;
--- a/src/server/live-reload/live-reload.js
+++ b/src/server/live-reload/live-reload.js
@@ -49,7 +49,7 @@ LiveReload.prototype._onFileChanged = function (fileRelativePath, parentDir) {
     if (this._forcePrepare) {
         // Sometimes, especially on Windows, prepare will fail because the modified file is locked for a short duration
         // after modification, so we try to prepare twice.
-        propagateChangePromise = retryAsync(this._project.prepare.bind(this._project), 2)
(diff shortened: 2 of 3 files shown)


src/server/utils/prepare.js
```diff
@@ -10,21 +10,43 @@ var exec = require('child_process').exec,
...
 
-    exec('cordova prepare ' + platform, {
-        cwd: projectRoot
-    }, function (err, stdout, stderr) {
-        if (err) {
-            deferred.reject(err || stderr);
+        exec('cordova prepare ' + platform, {
+            cwd: projectRoot
+        }, function (err, stdout, stderr) {
+            if (err || stderr) {
```
MSLaguana: Do we want to fail the build even if there is just a warning printed to stderr? Is Cordova guaranteed to only use stderr for actual errors, and not for warnings?

src/server/utils/prepare.js
```diff
@@ -10,21 +10,43 @@ var exec = require('child_process').exec,
  * @return {Promise}
  */
 function execCordovaPrepare(projectRoot, platform) {
-    var deferred = Q.defer();
+    return retryAsync(getExecCordovaPrepareImpl(projectRoot, platform), 2);
```
albertinad: By reading the name of the parameters from the retryAsync function, the number `2` would be the number of attempts to retry. Maybe we can avoid hardcoding this value here and define a constant at module level like,
{code}
var DEFAULT_RETRY_ATTEMPTS = 2;
{code}

src/server/utils/prepare.js
```diff
@@ -10,21 +10,43 @@ var exec = require('child_process').exec,
...
+            } else {
+                deferred.resolve();
+            }
+        });
+
+        return deferred.promise;
+    };
+}
+
+function retryAsync(promiseFunc, maxTries, delay, iteration) {
```
albertinad: We can improve the documentation here, at least of the parameters, to know which are required and which are optional.

src/server/utils/prepare.js
TimBarham: I vaguely remembered a scenario where the reverse was true - no exception was thrown and we only got info about an error through `stderr` rather than `err`. However, since that change isn't needed for this fix, to be safer I'll pull it until we have definite evidence it is needed (and safe).

src/server/utils/prepare.js
TimBarham: Sure. I didn't do anything like that because I was essentially just moving existing code. But no harm in making it clearer :). In fact I might move the `retryAsync` method to `jsUtils` and make 2 retries *it's* default.

src/server/utils/prepare.js
MSLaguana: For what it's worth, the time that I saw this happen to me the error message was actually in stdout, as well as returning a nonzero error code and thus populating `err`

src/server/utils/prepare.js
TimBarham: As above.

src/server/utils/prepare.js
albertinad: Thank you @TimBarham ! I forgot to mention that these changes are not stopper for merging this change :)
I think is great to move the `retryAsync` function to `jsUtils`.

albertinad:
This looks good to me! Just minor comments that should not block merging this change.

.eslintrc.json
```diff
@@ -5,6 +5,7 @@
         "commonjs": true
     },
     "rules": {
+        "no-unused-vars": ["error", {"vars": "all", "args": "none"}],
```
TimBarham: What do you think, @albertinad and @MSLaguana? Personally I don't like having a restriction against unused parameters, because sometimes they're useful (particularly on callback functions -  makes it clearer in the code the parameters the functions the code will be called with). Our jshint config allowed for this, but not our eslint config.

MSLaguana: I would prefer if we were explicit in when an argument is intended to be unused, for example doing `function (foo, _bar) {/* things using foo but not bar */ }`. It looks like eslint supports that via http://eslint.org/docs/2.0.0/rules/no-unused-vars#ignore-identifiers-that-match-specific-patterns

albertinad: Agree with @MSLaguana

Issue #114, opened 2016-07-13
Camera: file picker doesn't validate that the result is a valid content


albertinad:
When choosing a file as a response of the exec call of the plugin, or configuring the default image to use as a response, the result of selecting or cancelling picking an image is not validated. When the user cancel, the content is undefined, and an exception is thrown.

```
Uncaught TypeError: Failed to execute 'createObjectURL' on 'URL': No function was found that matched the signature provided.
```

albertinad:
@TimBarham maybe we should include this fix as part of the new release. Sorry to merge it late!

albertinad:
Fixed in #116

TimBarham:
It's ok - I mainly want the `0.2.x` branch to quickly fix an issue where `cordova-simulate` can completely fail to launch from VS Code. This camera plugin fix will be in our next release from master, and I don't expect it to be too long before we update VS Code to use that.

albertinad:
Sounds good!

PR #116, opened 2016-07-14
Updating camera plugin simulation image file handling.

albertinad:
This PR fixes thei ssue #114 and makes a few changes in the workflow. 
Summary of changes:
- Handling cases in which the user cancel the file picker selection
- Keep the selected files in variables, to not depend on the current files in the input file element. So when a user has a previous file selected, and when selecting a new one, it cancels, the file picker (OS dialog), then he could use the previous selected file, if any, for the current selection type.
- Removing commented code, that was disabled until those options were fully supported as part of the simulation. When we support more camera selection options, then we can review the history for the file, and take back that code if we still need it.

Thanks!

### DIFF
--- a/src/plugins/cordova-plugin-camera/sim-host.js
+++ b/src/plugins/cordova-plugin-camera/sim-host.js
@@ -9,12 +9,13 @@ var baseProps = {
 };
 
 module.exports = function (messages) {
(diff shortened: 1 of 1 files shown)


albertinad:
@MSLaguana @TimBarham is there any update on this? thanks! :)

TimBarham:
:+1:

PR #119, opened 2016-07-20
Move prepare retry logic from LiveReload to execCordovaPrepare.

TimBarham:
This is a version of the same change as before, but for the 0.2.x branch (underlying code has changed enough that I can't just cherry pick this one).


albertinad:
LGTM!

PR #121, opened 2016-07-22
Use send-transform npm package instead of tbarham/send (0.2.x version)

TimBarham:
This PR is against the 0.2.x branch.

I created a `send-transform` npm package. Version 0.13.0 is the same code as we were using from my github branch, and I'm applying this to the 0.2.x `cordova-simulate` branch because I want to use it in `vscode-cordova`. I'll make an equivalent change to master that uses version 0.14.1, which is based on the most recent `send` code.


albertinad:
LGTM!

PR #122, opened 2016-07-22
Use send-transform npm package instead of tbarham/send

TimBarham:
I created a `send-transform` npm package. Version 0.14.1 which I'm including here is based on the most recent `send` code.


albertinad:
LGTM!

PR #124, opened 2016-08-04
Improving XHR proxy

### LINKED ISSUE #111: XHR proxying omits static properties
When we enable XHR proxying, the XMLHttpRequest object no longer has any of its static properties, such as `DONE`.

### LINKED ISSUE #112: XHR proxy can't handle all failures correctly
If an app attempts to make an XHR which fails, we currently don't report the failure back to the client. By failure here I don't mean a 400 or a 500 error, but rather a non-existent domain, or a connection failed or reset. Currently these errors result in cordova-simulate crashing, but if we detect these errors and avoid crashing we still won't be able to report them back to the app in a transparent manner since the call from app_host into cordova-simulate has already successfully connected.

MSLaguana:
With the XHR proxy enabled, previously we would lose the static properties of XMLHttpRequest, such as DONE and LOADING. We now make sure to include these properties.

Additionally, since the connection from the app to the proxy always succeeds we do not correctly handle network level failures. Now if there is an error in the communication itself (as distinct from recieving a 404 for example) we report back a 502 Bad Gateway error, and include the error message to inform the app developer about the issue.

fixes #111
fixes #112

### DIFF
--- a/src/modules/xhr-proxy.js
+++ b/src/modules/xhr-proxy.js
@@ -24,6 +24,9 @@ var init = function () {
 
         return xhr;
     };
+
+    // Ensure that we retain all static properties, such as DONE
+    window.XMLHttpRequest.__proto__ = _XMLHttpRequest;
 };
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


albertinad:
Looks good to me!

PR #127, opened 2016-08-05
Fixing race condition in subscribing to socket events

MSLaguana:
When using VSCode I found that a single `cordova.exec` call from APP_HOST would sometimes trigger two responses from SIM_HOST. This turns out to be because `onSimHostReady` could be called twice without `resetAppHostState` called in between. As a fix, this commit ensures that each handler is only registered once.

### DIFF
--- a/src/server/socket.js
+++ b/src/server/socket.js
@@ -309,6 +309,7 @@ SocketServer.prototype._subscribeTo = function (host, msg, handler, once) {
     var socket = this._hostSockets[host],
         method = once ? 'once' : 'on';
     if (socket) {
+        socket.removeListener(msg, handler);
         socket[method](msg, handler);
     } else {
         log.log('Subscribing to a disconnected ' + host + ' wanting \'' + msg + '\'');


src/server/socket.js
```diff
@@ -309,6 +309,7 @@ SocketServer.prototype._subscribeTo = function (host, msg, handler, once) {
     var socket = this._hostSockets[host],
         method = once ? 'once' : 'on';
     if (socket) {
+        socket.removeListener(msg, handler);
```
TimBarham: I wonder if this should be a more targeted fix rather than doing this for every `_subscribeTo` call. In general, this method should support multiple subscribers to the same message. 

@busykai - what do you think?

MSLaguana: Note that I'm also specifying the handler to remove as well, so it should only catch repeated subscriptions with the same function, not multiple different subscriptions to the same message.

TimBarham: Talked it over with @MSLaguana, and agree this looks fine to me. Couple of things:

* Would still be good if you could take a look @busykai.
* I'm concerned this is just a symptom of a bigger problem. It would be good if the three of us could get together and talk this through at some point (perhaps @MSLaguana you could send an email describing the scenario in which this occurs).

albertinad:
@MSLaguana @TimBarham  I will review this on Monday.

albertinad:
@MSLaguana I've tried to reproduce this behaviour on latest master, and I wasn't able to reproduce the behaviour you described. VSCode extension is using `0.2.x` branch, please correct me if I'm wrong. I will try with that branch too.

MSLaguana:
Yeah, we were using the `0.2.x` branch. I've found it difficult to reproduce by hand, but it is a scenario that vs code seems to run into very easily when you simulate twice in a row.

albertinad:
I haven't tried yet on branch `0.2.x`, but now that you mention simulate twice in a row, probably the clean up after stopping is not done well in the previous implementation. There is a difference,
with the re-design, the API to reset the state after stopping looks like [this](https://github.com/Microsoft/cordova-simulate/blob/master/src/server/socket.js#L127):
```
SocketServer.prototype.closeConnections = function () {
    // stop watching file changes
    if (this._simulatorProxy.config.liveReload) {
        this._liveReload.stop();
    }

    Object.keys(this._hostSockets).forEach(function (hostType) {
        var socket = this._hostSockets[hostType];
        if (socket) {
            socket.disconnect(true);
        }
    }.bind(this));

    // not need to close the server, since it is closed
    // by the SimulationServer instance
    this._io = null;
    this._hostSockets = {};
    this._pendingEmits = {};
    this._pendingEmits[APP_HOST] = [];
    this._pendingEmits[SIM_HOST] = [];
    this._pendingEmits[DEBUG_HOST] = [];
    this._whenAppHostConnected = Q.defer(); // this is difference from 0.2.x
    this._whenSimHostReady = Q.defer(); // this is difference from 0.2.x

```

and in `0.2.x` looks like [this](https://github.com/Microsoft/cordova-simulate/blob/0.2.x/src/server/socket.js#L30):
```
function reset() {
    hostSockets = {};
    pendingEmits = {};
    pendingEmits[APP_HOST] = [];
    pendingEmits[SIM_HOST] = [];
    pendingEmits[DEBUG_HOST] = [];
    // reseting the promises is missing, and it's never done
```

If you add the following on that function,
```
whenAppHostConnected = Q.defer();
whenSimHostReady = Q.defer();
```
You can still reproduce it in VSCode?

albertinad:
Finally I was able to reproduce this behaviour in both branches: master and 0.2.x, even with the fix I shared above. I will continue taking a look at this.

albertinad:
@MSLaguana I couldn't continue to finding the root cause to this issue, so if you  need the, fix, you can merge it, and then continue doing some test to find out the root cause behind this issue and make a proper fix to that.
Thanks!

MSLaguana:
Thanks for looking into this @albertinad!

PR #129, opened 2016-08-09
Updating sim hosts file injection while streaming app host.

### LINKED ISSUE #128: App is not simulated when <head> contains any data-* attribute
When adding data-* attributes to <head> element from the index.html, simulation files are not injected, so cordova loose the simulated native bridge and enter in an infinite loop of processing messages.

screen shot 2016-08-09 at 2 18 02 pm

screen shot 2016-08-09 at 2 18 27 pm

albertinad:
Update regex to support matching `<head>` with content like `data-*` attributes. 
Also updating replace mechanism with [captures](https://github.com/eugeneware/replacestream/#search-and-replace-with-regular-expressions) for inserting the new head: do not replace it
with a new one, take the current head and add the script files.
This fixes #128.

Thanks!

### DIFF
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -223,7 +223,7 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
             send(request, filePath, {
                 transform: function (stream) {
                     return stream
-                        .pipe(replaceStream(/<\s*head\s*>/, '<head>' + scriptTags))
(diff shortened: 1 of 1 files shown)


MSLaguana:
Looks good to me

TimBarham:
Thanks! :shipit:

albertinad:
Thank you! merged.

PR #131, opened 2016-08-09
Changing to semantic CSP injection

### LINKED ISSUE #130: Specifying a connect-src in the CSP breaks debugging
As reported in https://github.com/Microsoft/vscode-cordova/issues/175 if an app specifies a connect-src in its CSP then the websocket connections to the simulation server will be rejected.

Rather than injecting `default-src 'self' ws:` a better solution may be to specifically add `connect-src 'self' ws:`

MSLaguana:
Instead of simply injecting some additional entries into the default-src of the csp, we now properly parse it and add the specific constraints that we care about. In particular, we make sure to preserve any previous constraints, but we allow websocket connections to 'self' and blob urls for images.

Fixes #130

### DIFF
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -15,6 +15,13 @@ var fs = require('fs'),
     pluginSimulationFiles = require('./plugin-files');
 
 /**
+ * Maximum length of <meta> tag we search for when modifying CSP properties
+ * Needed for when a tag crosses multiple chunk boundaries.
+ * @const
+ */
+var MAX_CSP_TAG_LENGTH = 1024;
+
+/**
  * The simulation server that encapsulates the HTTP server instance and
  * Web Socket server. It enables to start and stop the server to listening
  * for new connections.
@@ -225,7 +232,6 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
             var metaTagRegex = /<\s*meta[^>]*>/g;
             var cspRegex = /http-equiv\s*=\s*(['"])Content-Security-Policy\1/;
             var cspContent = /(content\s*=\s*")([^"]*)"/;
-            var maxCspTagLength = 1024;
             send(request, filePath, {
                 transform: function (stream) {
                     return stream
@@ -248,7 +254,7 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
                                 policy.add('img-src', 'blob:');
                                 return preamble + policy.toString() + '"';
                             });
-                        }, {maxMatchLen: maxCspTagLength}));
+                        }, {maxMatchLen: MAX_CSP_TAG_LENGTH}));
                 }
             }).pipe(response);
         }.bind(this))
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -4,6 +4,7 @@ var fs = require('fs'),
     path = require('path'),
     replaceStream = require('replacestream'),
     cordovaServe = require('cordova-serve'),
+    cspParse = require('csp-parse'),
     send = require('send-transform'),
     url = require('url'),
     Q = require('q'),
@@ -218,13 +219,35 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
                 return '<script src="' + scriptSource + '"></script>';
             }).join('');
 
-            // Note we replace "default-src 'self'" with "default-src 'self' ws:" (in Content Security Policy) so that
-            // websocket connections are allowed (this relies on a custom version of send that supports a 'transform' option).
+            // Note we add "connect-src 'self' ws:" and "img-src 'blob:' (in Content Security Policy) so that
+            // websocket connections are allowed and the camera plugin works (this relies on a custom version
+            // of send that supports a 'transform' option).
+            var metaTagRegex = /<meta[^>]*>/;
+            var cspRegex = /http-equiv\s*=\s*(['"])Content-Security-Policy\1/;
+            var cspContent = /(content\s*=\s*")([^"]*)"/
             send(request, filePath, {
                 transform: function (stream) {
                     return stream
                         .pipe(replaceStream(/(<\s*head[^>]*>)/, '$1' + scriptTags))
-                        .pipe(replaceStream('default-src \'self\'', 'default-src \'self\' ws: blob:'));
+                        .pipe(replaceStream(metaTagRegex, function (metaTag) {
+                            if (!cspRegex.test(metaTag)) {
+                                // Not a CSP tag; return unchanged
+                                return metaTag;
+                            }
+                            return metaTag.replace(cspContent, function (match, preamble, csp) {
+                                var policy = new cspParse(csp);
+                                var defaultCsp = policy.get('default-src');
+                                if (!policy.get('connect-src')) {
+                                    policy.add('connect-src', defaultCsp);
+                                }
+                                policy.add('connect-src', "'self' ws:");
+                                if (!policy.get('img-src')) {
+                                    policy.add('img-src', defaultCsp);
+                                }
+                                policy.add('img-src', 'blob:');
+                                return preamble + policy.toString() + '"';
+                            });
+                        }));
                 }
             }).pipe(response);
         }.bind(this))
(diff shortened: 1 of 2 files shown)


src/server/server.js
```diff
@@ -218,13 +219,35 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
                 return '<script src="' + scriptSource + '"></script>';
             }).join('');
 
-            // Note we replace "default-src 'self'" with "default-src 'self' ws:" (in Content Security Policy) so that
-            // websocket connections are allowed (this relies on a custom version of send that supports a 'transform' option).
+            // Note we add "connect-src 'self' ws:" and "img-src 'blob:' (in Content Security Policy) so that
+            // websocket connections are allowed and the camera plugin works (this relies on a custom version
```
albertinad: What do you mean that the camera plugin works? Is there any issue with that?

src/server/server.js
```diff
@@ -218,13 +219,35 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
...
+                            }
+                            return metaTag.replace(cspContent, function (match, preamble, csp) {
+                                var policy = new cspParse(csp);
+                                var defaultCsp = policy.get('default-src');
+                                if (!policy.get('connect-src')) {
+                                    policy.add('connect-src', defaultCsp);
+                                }
+                                policy.add('connect-src', '\'self\' ws:');
+                                if (!policy.get('img-src')) {
+                                    policy.add('img-src', defaultCsp);
```
albertinad: Maybe we can add the comment related to the camera plugin and image handling at this point.

src/server/server.js
```diff
@@ -218,13 +219,35 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
...
+                                }
+                                policy.add('img-src', 'blob:');
+                                return preamble + policy.toString() + '"';
+                            });
+                        }, {maxMatchLen: 1024}));
```
albertinad: nit: we can keep `maxMatchLen` configuration value in a constant, with a representative name and maybe a comment that clarifies the reason of that specific value, so we don't forget about it.

src/server/server.js
```diff
@@ -218,13 +219,35 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
...
 
-            // Note we replace "default-src 'self'" with "default-src 'self' ws:" (in Content Security Policy) so that
-            // websocket connections are allowed (this relies on a custom version of send that supports a 'transform' option).
+            // Note we add "connect-src 'self' ws:" and "img-src 'blob:' (in Content Security Policy) so that
+            // websocket connections are allowed and the camera plugin works (this relies on a custom version
+            // of send that supports a 'transform' option).
+            var metaTagRegex = /<\s*meta[^>]*>/;
+            var cspRegex = /http-equiv\s*=\s*(['"])Content-Security-Policy\1/;
+            var cspContent = /(content\s*=\s*")([^"]*)"/
             send(request, filePath, {
```
albertinad: nit: The call to `send` performs a critical task in order to make the simulation works: inject simulation files and makes an analyzisis of CSP in order to inject another configuration that allows connections to localhost for simulation. We could probably do a little refactor on `_streamAppHostHtml` and the promise chain at the prepare call. So it is easy to follow, smaller methods and more testable.
This is not needed to do it as part of this PR, it's just a comment :)

src/server/server.js
```diff
@@ -218,13 +219,35 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
...
             }).join('');
 
-            // Note we replace "default-src 'self'" with "default-src 'self' ws:" (in Content Security Policy) so that
-            // websocket connections are allowed (this relies on a custom version of send that supports a 'transform' option).
+            // Note we add "connect-src 'self' ws:" and "img-src 'blob:' (in Content Security Policy) so that
+            // websocket connections are allowed and the camera plugin works (this relies on a custom version
+            // of send that supports a 'transform' option).
+            var metaTagRegex = /<\s*meta[^>]*>/;
+            var cspRegex = /http-equiv\s*=\s*(['"])Content-Security-Policy\1/;
+            var cspContent = /(content\s*=\s*")([^"]*)"/
```
albertinad: Missing semi-colon

src/server/server.js
```diff
@@ -218,13 +219,35 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
                 return '<script src="' + scriptSource + '"></script>';
             }).join('');
 
-            // Note we replace "default-src 'self'" with "default-src 'self' ws:" (in Content Security Policy) so that
-            // websocket connections are allowed (this relies on a custom version of send that supports a 'transform' option).
+            // Note we add "connect-src 'self' ws:" and "img-src 'blob:' (in Content Security Policy) so that
+            // websocket connections are allowed and the camera plugin works (this relies on a custom version
+            // of send that supports a 'transform' option).
+            var metaTagRegex = /<\s*meta[^>]*>/;
```
albertinad: I've tested these changes with the cordova mobilespec app, but with a little modification. The cordova mobilespec app analyzes the user agent to identify mobile runtime, and if it matches some of the OS like android, ios etc, it will inject CSP. That is done automatically when one of the scripts is loaded. Since for cordova-simulate with do  not change the user agent yet, I took the CSP configuration and add it manually to the index.html, looking like this:
```
<html>
  <head>
    <!-- meta tag #1 -->
    <meta name="viewport" content="width=device-width,height=device-height,user-scalable=no,initial-scale=1.0" />
    <!-- meta tag #1 -->
    <meta http-equiv="Content-type" content="text/html; charset=utf-8"/>
    <!-- meta tag #3 - CSP -->
    <meta http-equiv="Content-Security-Policy" content="default-src 'self' https://ssl.gstatic.com/accessibility/javascript/android/; connect-src 'self' http://cordova-filetransfer.jitsu.com; media-src 'self' http://cordova.apache.org/downloads/ https://cordova.apache.org/downloads/; frame-src 'self' http://stealbridgesecret.test/ data: gap:; img-src 'self' data: cdvfile:; style-src 'self' 'unsafe-inline'"/>
    <!-- more content here -->
 
  </head>
  <body id="stage" class="theme">
    <!-- more content here -->
  </body>
</html>
```

With the changes in this branch, and also with latest master, replacing CSP while streaming the index.html doesn't work. The regex to match the meta tag only matches the first meta tag, not all of them. But if we change the regex to use the global modifier,
 ```
var metaTagRegex = /<\s*meta[^>]*>/g;
```
it returns all of the matches.

File: src/server/server.js (hunk 1 above)
Comment (MSLaguana) [inline src/server/server.js]: My understanding is that the camera plugin works by providing `blob:` URLs, and that's the reason for adding it to the CSP.

File: src/server/server.js (hunk 6 above)
Comment (MSLaguana) [inline src/server/server.js]: Ah good point! I'll make the change.

File: src/server/server.js (hunk 1 above)
Comment (albertinad) [inline src/server/server.js]: Thanks for the clarification.

File: src/server/server.js (hunk 5 above)
Comment (MSLaguana) [inline src/server/server.js]: Fixed

File: src/server/server.js [hunk 7]
```diff
@@ -218,13 +219,36 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
...
 
-            // Note we replace "default-src 'self'" with "default-src 'self' ws:" (in Content Security Policy) so that
-            // websocket connections are allowed (this relies on a custom version of send that supports a 'transform' option).
+            // Note we add "connect-src 'self' ws:" and "img-src 'blob:' (in Content Security Policy) so that
+            // websocket connections are allowed and the camera plugin works (this relies on a custom version
+            // of send that supports a 'transform' option).
+            var metaTagRegex = /<\s*meta[^>]*>/g;
+            var cspRegex = /http-equiv\s*=\s*(['"])Content-Security-Policy\1/;
+            var cspContent = /(content\s*=\s*")([^"]*)"/;
+            var maxCspTagLength = 1024;
```
albertinad: Maybe this could be declared only once at module level,
```
/**
 * Some doc if we need to clarify something.
 * @const
 */
var MAX_CSP_TAG_LENGTH = 1024;
```

albertinad:
Looks good to me! thanks!

PR #132, opened 2016-08-10
Rename browserify/app-host global require.

albertinad:
Rename the `require` global variable injected by app-host (browserify) to another name, avoiding conflicts with other module loaders like `requirejs`.
Intel XDK user [reported issue](https://software.intel.com/en-us/forums/intel-xdk/topic/673464).

### DIFF
--- a/src/server/sim-files.js
+++ b/src/server/sim-files.js
@@ -142,7 +142,8 @@ SimulationFiles.prototype._createHostJsFile = function (simulationFilePath, host
     var b = browserify({
         paths: getBrowserifySearchPaths(hostType),
         debug: true,
-        externalRequireName: '_cordovaSimulateRequire'
+        exports: false, // needed to prevent browserify to override hasExports option by set it to true
+        hasExports: false
     });
 
     b.transform(function (file) {
--- a/src/server/sim-files.js
+++ b/src/server/sim-files.js
@@ -139,7 +139,12 @@ SimulationFiles.prototype._createHostJsFile = function (simulationFilePath, host
 
     var scriptDefs = createScriptDefs(hostType, scriptTypes);
 
-    var b = browserify({ paths: getBrowserifySearchPaths(hostType), debug: true });
+    var b = browserify({
+        paths: getBrowserifySearchPaths(hostType),
+        debug: true,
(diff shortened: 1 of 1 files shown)


src/server/sim-files.js
```diff
@@ -139,7 +139,12 @@ SimulationFiles.prototype._createHostJsFile = function (simulationFilePath, host
 
     var scriptDefs = createScriptDefs(hostType, scriptTypes);
 
-    var b = browserify({ paths: getBrowserifySearchPaths(hostType), debug: true });
+    var b = browserify({
+        paths: getBrowserifySearchPaths(hostType),
+        debug: true,
+        externalRequireName: '_cordovaSimulateRequire'
```
albertinad: `externalRequireName` is an option passed to `browser-pack`, [more info](https://github.com/substack/browser-pack#packopts).

TimBarham:
Does the resulting pack actually need to expose an external `require`?

src/server/sim-files.js
MSLaguana: Looks like `opts.externalRequireName` is only used if `opts.hasExports` is specified, which we don't. Although looking closer at browserify, it does seem to specify `hasExports` if `exports` is undefined and `opts.entry` is falsy, which looks to be the case at the moment.

TimBarham: Yeah, that's what I'm wondering - if the proper fix is to just explicitly set `hasExports` to `false` (or whatever we need to do to make it `false`).

albertinad: I passed that option, `hasExports` to false, instead of `externalRequireName`, and nothing happen, `require` is still exports to the global. I will research this.

albertinad:
No, I don't think that the resul pack needs to export a any variable to the global scope. Let me propose some alternatives to this issue.

src/server/sim-files.js
albertinad: @MSLaguana I will try that combination of options.

TimBarham:
:shipit:

MSLaguana:
Looks good to me

albertinad:
Great! thank you!

PR #133, opened 2016-08-16
Support simulation when there's no plugins added to the project.

albertinad:
There are some cases during the development of a cordova app in which developers  doesn't add any plugins to the project. They even remove the whitelist plugin, that is included when adding a platform to the project. 
When trying to simulate an app without any plugin, the server crashes due to an ENOENT exception, 
```
SIM: Server started:
- App running at: http://localhost:8000/index.html
- Sim host running at: http://localhost:8000/simulator/index.html
/Users/<user>/cordova-simulate/node_modules/q/q.js:155
                throw e;
                      ^
Error: ENOENT, open '/Users/<user>/apps/device/plugins/android.json'
    at Error (native)

```
The platform json file is not generated, it doesn't exists.
Once that is fixed, when removing all the plugins and running cordova prepare and start the simulation, device ready is never fired due to an unhandled exception in the execution of the simulation protocol.
```
cordova.js:59 Uncaught module cordova/plugin_list not found
```
This happens when doing
```
cordova.require('cordova/plugin_list')
```
to get the list of installed plugins. The `cordova/plugin_list` module is loaded and available when the `cordova_plugins.js` file is successfully created as a result of the cordova prepare operation. When no plugins are available, that file is not created, and the module is not loaded.

This PR fixes this corner cases in which a developer simulates the app that doesn't have any plugin installed.

### DIFF
--- a/src/server/project.js
+++ b/src/server/project.js
@@ -70,12 +70,6 @@ Project.DEFAULT_PLUGINS = [
     'events'
 ];
 
-/**
- * @const
- * @private
- */
-Project._PLUGINS_LIST_BAD_STATE = ['__unknown__'];
-
 Project.prototype.initPlugins = function () {
     this._resetPluginsData();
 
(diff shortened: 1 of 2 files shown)


src/server/project.js
```diff
@@ -317,6 +326,11 @@ Project.prototype._getProjectState = function() {
             var pluginsJsonPath = path.join(projectRoot, 'plugins', platform + '.json');
             return Q.nfcall(fs.readFile, pluginsJsonPath);
         })
+        .fail(function () {
+            // an error ocurred trying to read the file for the current platform,
+            // set the pluginsList to indicate a "bad state".
+            newState.pluginList = Project._PLUGINS_LIST_BAD_STATE;
```
MSLaguana: I think that this isn't quite right; since you don't rethrow an error this will pass void to the next promise handler, which will then cause the `fileContent.toString()` to fail and go to the error handler, but I'm not sure what the point of directly assigning to `newState.pluginList` here is.

I'd prefer to either `return '{}';` to convert a missing file into an empty set of installed plugins, or to put this failure handler after the next then so we only assign to `newState.pluginList` once.

albertinad: You're right, this is wrong. I will change it.

MSLaguana:
Looks good to me now!

albertinad:
Thanks! merged!

PR #134, opened 2016-08-22
added bin

edewit:
To enable to run the command from the command line added 'bin' section

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,18 @@ npm install -g cordova-simulate
 
 # Usage
 
+## CLI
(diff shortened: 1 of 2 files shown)


TimBarham:
Thanks @edewit! We'll take this PR. But while you're at it, it would be great if you could restore the documentation for command line usage in `README.md` that was removed when bin was removed - would be nice to have that as part of the same commit. The commit where those docs were removed is here: Microsoft/taco-simulate@9d80ce15.

edewit:
@TimBarham cool, added cli usage, to the readme

PR #139, opened 2016-08-23
Rendering improvements in non-Chrome browsers.

TimBarham:
Chrome renders columns differently to other browsers. The current CSS was optimized for Chrome, and didn't work well in IE (top of columns wasn't aligned, often had empty first column). This change tweaks some values to be specific for Chrome and some values only for non-Chrome browsers (the non-Chrome values also work fine in Firefox).

I'll keep working towards an approach that works consistently across browsers without needing special case stuff, but until I get there, I need this change to unblock current work.

### DIFF
--- a/src/sim-host/ui/server/server.js
+++ b/src/sim-host/ui/server/server.js
@@ -4,22 +4,35 @@ var path = require('path'),
     replaceStream = require('replacestream'),
     send = require('send-transform');
 
+var BROWSER_REPLACE_MAX_MATCH_LENGTH = 1024;
+var BROWSER_REPLACE_OPTIONS = {maxMatchLen: BROWSER_REPLACE_MAX_MATCH_LENGTH};
+
 module.exports.attach = function (app, dirs, hostRoot) {
-    app.get('/simulator/sim-host.css', function (request, response, next) {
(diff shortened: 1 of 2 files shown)


src/sim-host/ui/server/server.js
```diff
@@ -5,21 +5,31 @@ var path = require('path'),
...
+            transform: getTransform(userAgent)
+        }).pipe(response);
     });
 };
+
+function getTransform(userAgent) {
+    if (isChrome(userAgent)) {
+        // If target browser is Chrome, remove any sections marked as not for Chrome.
+        return function (stream) {
+            return stream.pipe(replaceStream(/\/\* BEGIN !CHROME \*\/[\s\S]*\/\* END \*\//gm, ''));
```
MSLaguana: I think that these regexes will be greedy, and remove from the first instance to the last.

TimBarham: Worked by coincidence that `replaceStream()` wasn't hitting the second end "tag". Bumped `maxMatchLen` up to 1024 for these matches, and made end "tags" unique.

MSLaguana:
Apart from exposing `maxMatchLen` as a global / configurable parameter, looks good to me.

PR #140, opened 2016-08-24
Configure "step" attribute of all cordova-input-number used.

### LINKED ISSUE #117: cordova-number-entry fields indicate valid values are invalid
Some browsers (Chrome for example) validate the value of a numeric input field based on the value of `step`. Since this defaults to 1, and we never specify it for `cordova-number-entry` elements, any non-integer value is flagged as an error. For example, the tooltip for the Latitude field of the Geolocation panel shows something like the following error:

    Please enter a valid value. The two nearest valid values are 43 and 44.

albertinad:
Not all of the `cordova-number-entry` elements were configured properly according to the plugin API specification. This PR is to update the attribute configuration of `cordova-number-entry`, mostly for the geolocation widget.
This fix #117

### DIFF
--- a/src/plugins/cordova-plugin-device-orientation/sim-host-panels.html
+++ b/src/plugins/cordova-plugin-device-orientation/sim-host-panels.html
@@ -34,7 +34,7 @@
         <cordova-panel-row>
             <label for="compass-heading-value">Heading (deg)</label>
             <label data-compass-heading="text">N</label>
-            <input type="number" id="compass-heading-value" class="cordova-state-default" min="0" max="359.99" step="0.5">
+            <input type="number" id="compass-heading-value" class="cordova-state-default" min="0" max="359.99" step="any">
         </cordova-panel-row>
     </cordova-group>
 </cordova-panel>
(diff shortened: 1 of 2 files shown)


MSLaguana:
Looks like we'll change from half unit steps to full unit on the orientation plugin, but it's still editable so I think that's fine.

Should we be imposing minimums on altitude? It is possible to be below sea level, such as at the dead sea.

Looks good in general.

albertinad:
@MSLaguana I follow this (documentation)[https://shapeshed.com/html5-geolocation-api/] to set the `min` and `step`.

TimBarham:
Altitude isn't spec'd as >= 0, is it? (seems negative numbers should be valid)

TimBarham:
Not needed for this PR, but we should provide a way to specify `null` for values that support it.

albertinad:
You're right @TimBarham, the [spec](https://www.w3.org/TR/geolocation-API/) doesn't specify anything in particular about the value that they can take, like the others. So, negative numbers seems supported too.
I've created an issue for addressing the update of `cordova-plugin-geolocation` simulation in order to support `null` and `NaN` values according to the spec: #140 

Thanks!

Issue #142, opened 2016-08-25
Test infrastructure


albertinad:
Define and implement initial test infrastructure:
- unit test
- integration test

PR #145, opened 2016-08-31
Draft protocol refactoring (server-side).

busykai:
Protocol is refactored using state machine instead of promises. See PR #127 for
the description of the issue (race condition when reloading both `SIM_HOST` and
`APP_HOST` simultaneously) and implemented workaround.

This is a draft, do not merge. Do not review in gory details. Just please take a quick look and let me know if you think that this is a cleaner implementation (it is, IMO) than the previous one. It's functional, except for some cases (e.g. `SIM_HOST` reload is not addressed properly yet). If we come to an agreement, this is better, I will finish it up rapidly.

CC: @MSLaguana, @TimBarham, @albertinad


src/server/socket.js
```diff
@@ -28,20 +44,112 @@ function SocketServer(simulatorProxy, project) {
...
+            this._setupSimHostHandlers();
+            this._subscribeTo(APP_HOST, 'app-plugin-list', function (data) {
+                this._emitTo(SIM_HOST, 'app-plugin-list', data);
+                this._setAppHostState(AH_STATE_PLUGINS_READY);
+            }.bind(this), true);
+            this._emitTo(APP_HOST, 'init');
+            this._setAppHostState(AH_STATE_INIT_SENT);
+            break;
+        case AH_STATE_PLUGINS_READY | SH_STATE_READY:
+            this._subscribeTo(SIM_HOST, 'start', function () {
```
MSLaguana: Does this introduce a new unintended synchronization point? Is it possible for sim-host to enter the `READY` state before app-host gets to the `PLUGINS_READY`state, so we don't enter this case here, and yet sim-host emits a `start` event?

MSLaguana:
I like the idea of the individual state machines for the different components, since the state is part of the protocol that we are using and it's a nice way of being explicit about it. I'm just not sure about the combination of the two together, and whether it makes it easier to miss asynchronous events if one component progresses faster or slower than the other.

src/server/socket.js
MSLaguana: Perhaps we should attach handlers to the sockets for all the events immediately, and have the handlers also check the state for what action should be taken, including possibly buffering output from one component to the other if they are out of sync

busykai: > Does this introduce a new unintended synchronization point?

This synchronization was [there](https://github.com/Microsoft/cordova-simulate/pull/145/files#diff-c97dea31357a0880834e2407a2dcbc97L238) already. This one is just an explicit expression of it.

> Is it possible for sim-host to enter the READY state before app-host gets to the PLUGINS_READYstate, so we don't enter this case here, and yet sim-host emits a start event?

It does not work this way. `_handleStateChange` is called each time either of the clients changes the state. So which ones gets to which state first does not matter. The complexity of the previous implementation was exactly what you describe: the timing. In the new scheme of things, the timing of the state changes does not matter and each combination of the states of the clients is processed once.

MSLaguana: The timing issues that I'm thinking of are not directly tied to the states as they stand. For example, suppose that app-host connects first, and so transitions to the `CONNECTED` state, but sim-host hasn't yet connected. Is it possible that app-host could independently fire the `app-plugin-list` event? Then when sim-host connects and transitions to `READY`, the above code will hook up a listener but it's too late because it's already fired?

busykai: No, it's not possible by the [protocol design](https://github.com/Microsoft/cordova-simulate/wiki/The-cordova-simulate-bootstrap-protocol) and by the implementation (to your specific point, `init` will never be sent to `APP_HOST` and `APP_HOST` will not send anything proactively until it's received `init`).

I was not probably explicit, but this code implements a spec. It is a built-in design point, though, that either of the clients could connect at any point in time. Could you review the [spec](https://github.com/Microsoft/cordova-simulate/wiki/The-cordova-simulate-bootstrap-protocol) and tell me if it's not detailed enough or if it has a flaw? It could use a fresh look. If the protocol is ok, we could then discuss the implementation. If it's flawed, it has to be fixed.

MSLaguana: Ah I see, that makes sense. Given the guarantees/assumptions of the protocol, this makes a lot of sense.

albertinad:
I really like this improvement based on state machines. This is a much more cleaner implementation, easy to follow and expand if it's necessary. We can probably iterate over some details to improve, but the overview looks good, and it seems the right approach to me.
About missing asynchronous events if some component progresses faster/slower than the other, this could be the first test case to have as part of #142 :)

PR #147, opened 2016-08-31
Adds ability to customize sim-host window title

TimBarham:


src/server/simulator.js
```diff
@@ -208,6 +208,7 @@ function parseOptions(opts) {
         /* use the default simulation UI */
         simHostOpts = { simHostRoot: path.join(__dirname, '..', 'sim-host', 'ui') };
     }
+    simHostOpts.title = opts.simhosttitle || "";
```
MSLaguana: Should we default to something instructive, like "<appname> Simulation"?

albertinad: I like that idea!

TimBarham: Sure :smile:

PR #149, opened 2016-09-14
If prepare fails, display a warning and continue.

TimBarham:
There are scenarios where Ripple will work but cordova-simulate won't (for example, ios platform with a new blank Ionic app). Rather than crashing the server, if prepare fails simply display a warning and continue.

### DIFF
--- a/src/server/project.js
+++ b/src/server/project.js
@@ -7,7 +7,8 @@ var fs = require('fs'),
     dirs = require('./dirs'),
     pluginUtil = require('./utils/plugins'),
     prepareUtil = require('./utils/prepare'),
-    utils = require('./utils/jsUtils');
+    utils = require('./utils/jsUtils'),
+    log = require('./utils/log');
 
 /**
  * The Project model encapsulates the information and state about the project under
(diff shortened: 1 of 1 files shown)


MSLaguana:
Looks good to me. I do feel like a failure to prepare is probably pretty bad, but the case you mention using ionic seems like reason enough to try to cope where possible.

src/server/project.js
```diff
@@ -192,9 +193,18 @@ Project.prototype.prepare = function () {
...
                         .then(function () {
                             return true;
+                        }).catch(function (err) {
+                            err = err.message || err.toString();
+                            var pos = err.toLowerCase().indexOf('error: ');
+                            if (pos > -1) {
+                                err = err.slice(pos + 7);
+                            }
+                            err = err.split('\n')[0];
+                            log.warning('Preparing platform \'' + platform + '\' failed: ' + err);
```
MSLaguana: Actually this error needs to return something. Right now it's returning void, while the success case is returning true. Is falsy enough here?

TimBarham: Yeah, fair point. Should return true here I believe.

PR #150, opened 2016-09-15
Updated Device panel and devices list.

TimBarham:
* Updates devices list to reflect modern devices.
* Updates Device panel to show more information and replaces Viewport panel.
* Allows simulated device to be specified via command line or options.
* Spoofs user agent based on platform, device and OS version.
* Allows different OS versions to be chosen for each device.
* Remembers last chosen device per platform (adds "local storage" support on the server side).

@busyka, @albertinad and @MSLaguana ... This is a pretty large change, and modifies some of the startup sequence, so I'd appreciate if you guys could take a good look at it. Also, this will certainly impact @busykai's current refactor work, and also how the XDK does platforms and devices and uses the Device panel.

Thanks!


src/server/server.js
```diff
@@ -162,30 +162,41 @@ SimulationServer.prototype._prepareRoutes = function () {
...
         hostJsFile = this._simulationFiles.getHostJsFile(hostType, config.simulationFilePath);
 
     if (!hostJsFile) {
         throw new Error('Path to ' + hostType + '.js has not been set.');
     }
 
-    response.sendFile(hostJsFile);
+    if (hostType === 'app-host') {
+        // Provide current user agent
+        send(request, hostJsFile, {
```
MSLaguana: I don't follow why this uses `request` rather than `response`. Can you even `send` to a HTTPRequest rather than a response?

MSLaguana: Oh nevermind, I had a brainfart. This is not sending to the request, it is using the request as a source, not sending to it.

src/sim-host/ui/sim-host.js
```diff
@@ -18,14 +20,17 @@ customElements.initialize(changePanelVisibilityCallback);
 
 window.addEventListener('DOMContentLoaded', function () {
     sizeContent();
-    Q.all([db.initialize(), initSocketPromise]).then(function () {
-        initializePlugins();
+    Q.all([db.initialize(), initSocketPromise]).then(function (result) {
```
MSLaguana: I'd suggest using `Q.all([...]).spread(function (initializeIgnored, initResult) {` for the extra documentation value.

TimBarham: In fact  think I could simply replace `Q.all()` with `Q.spread()`. The reason I didn't is that `spread()` is Q specific, and not part of the Promise standard. While we currently use Q, I think it's preferable to work within the Promise standard if possible (which is why I also tend to avoid `Q.defer` these days).

src/server/server.js
TimBarham: 😄

albertinad:
I will review this PR @TimBarham !
I didn't take a deeper look at all of the changes yet, but I noticed that you are implementing device selection in the `cordova-plugin-device`, and also loading it by default, like exec and geolocation, no matter if the user doesn't have that plugin in the project. I think that we can improve this. From our experience in the XDK emulator based on Ripple, the fact that we used the same simulation widget for: switching devices, show data related to this device like user agent and viewport size, and also showing the state of the `cordova-plugin-device`, this cause confusions to our users about what is being simulated and what plugins do they have in their projects.
I suggest a different approach. 
- Leave the `cordova-plugin-device` simulation to simulate the state of the real plugin, and all the metadata, and nothing more. Remove the device selector, since that is something you can't do with the cordova plugin, and just reflect the state. Do not always load this simulation control as you are doing in this PR, only load it when this plugin is added to the project.
- Create a new simulation control that has a list of supported devices, enables to select a device, displays information about it: viewport size, pixel ratio, user agent being used etc. Always load this simulation control as we do it with `exec`, `events` and `geolocation`.
The benefits of this approach is that we have a clear separation of functionality: `cordova-plugin-device` state, and device switching/simulation. Developers will not be confused about having device information done by `cordova-plugin-device` that is not present in the project but it is being simulated, and if they don't have device plugin in their projects, they still can simulate devices by changing them. The state and data for both simulation controls is clear, developers will not be confused about accessing some data like viewport size or pixel ratio or user agent through device plugin, because it doesn't exist such API in the plugin, is something we've added as part of device simulation, not `cordova-plugin-device` simulation
This new simulation control is a kind of special control, that is not related to any configuration of a plugin, or plugin presence in the project, but a special control that we use and reuse all the infraestructure for simulation.
The missing part here is that we need a way to make the `cordova-plugin-device` simulation control to listen events when a device is selected in the other control, so it displays the information that the plugin exposes. We have this infraestructure today, we can reuse and improve it.

We have implemented this approach in the Intel XDK, we have separated `cordova-plugin-device` simulation and device simulation into 2 simple cohesive simulation controls. The first one listens to events when a device is selected in other side, and get the information to display.
 
What do you think about this?

TimBarham:
Albertina - thanks for the great feedback! I've raised your feedback with the team and I'm sure we'll talk about it. In the meantime, here was my thinking for the approach found here.

Before implementing this change, I looked at what you guys had done in the XDK, and also what Ripple does. 

I went with the single panel for the following reasons:

* This is really all information about the current device - all things that change when you select different devices. It just makes sense to me that this all belongs in a single panel.
* In the current XDK implementation, there is duplicate information between the two panels. Also the two panels are physically "distant" from each other. Two things I'd prefer to avoid.
* Ripple doesn't have a Device panel that is only present when the device plugin is in the app - it just has a Devices panel and an Information panel, both of which are always present. Our existing customers are used to this, so I don't think there will be any confusion about having this Device panel always present. This implementation is essentially just a combination of those two Ripple panels (plus *some* information that is, admittedly, only really relevant if the device panel is present).

Some thoughts I've had:

* Rename this panel to "Device Information"
* *Possibly* remove some information that is device plugin specific and not very useful to display here even when the device plugin is present (like "UUID", "Version" and "Serial").
* Alternatively, only show those options when the device plugin is present.

PR #151, opened 2016-09-16
Fix live reload failing when launched from VS.

TimBarham:
VS generates some temporary files when saving files. Update live reload to ignore them.


src/server/utils/jsUtils.js
```diff
@@ -197,10 +197,20 @@ function retryAsync(promiseFunc, maxTries, delay, iteration) {
     });
 }
 
+function stripErrorColon(err) {
```
albertinad: If we need to later remove the 'error:' string, then probably we could not add it as part of the error message.

albertinad:
Looks good to me! just a minor comment, it should not stop merge this.

src/server/utils/jsUtils.js
TimBarham: Hmmm... I'm not sure what you mean. This is for processing errors that come from elsewhere and often have "Error:" at the start.

albertinad: You're right! I don't know what I commented this, I confused logging with throwing an error, I read above that we log an error with 'Error' but it is a log, not throwing an error. Just a problem with my brain, forget this :)

TimBarham: 😄

sgulley2:
I am installed cordova-simulator -g standalone outside visual studio because I noticed that live reload is not working when launched from vs2017 and it still does not work.  So live reload is not working for ionic 2 priojects either inside vs2017 or standalone

PR #153, opened 2016-09-26
Adds support for localization.

TimBarham:
* Adds `tools/i18n` folder with dev support scripts (`update.js` is the driving script).
* Adds `update-loc` script to `package.json` (which calls `update.js`).
* See `tools/i18m/README.md` for more information.

As described in the readme, `update.js` creates and maintains `xliff` files under the `tools/i18n directory`. Initially it populates these with machine generated translations (to make testing easier), but flags the loc team that they need real translations.

It also generates `*.i18n.json` files in an` i18n` directory under the `src` directory, which are used to replace strings in files as they are served up.


TimBarham:
Note that once this is finalized and merged I will open a second PR that contains the generated `xliff` and `i18n.json` files.

src/plugins/exec/sim-host-dialogs.html
```diff
@@ -4,8 +4,7 @@
     <cordova-group>
         <p>There is no handler for the following exec call:</p>
         <p style="font-style:italic;color:black"><span id="exec-service">service</span>.<span id="exec-action">action</span></p>
-        <p>Please enter a value below to be returned for this call, then indicate whether this represents success or failure.
-            If you wish to always respond to this call with this result, select the 'persist' option.</p>
+        <p>Please enter a value below to be returned for this call, then indicate whether this represents success or failure. If you wish to always respond to this call with this result, select the 'persist' option.</p>
```
MSLaguana: Why the change here? Was this an autoformat change, or are there restrictions on how newlines interact with loc?

src/server/server.js
```diff
@@ -394,4 +404,58 @@ function processPluginHtml(html, pluginId) {
     });
 }
 
+function localizeHtmlFile(htmlFile, html, lang) {
+    // Replace localized strings
+    var targetJsonFile = getI18nJsonFile(htmlFile, lang);
+    if (utils.existsSync(targetJsonFile)) {
+        try {
+            var locItems = require(targetJsonFile);
```
MSLaguana: I believe that this declaration should be moved out of the try, at least to avoid confusion when it is used further down.

src/server/server.js
```diff
@@ -394,4 +404,58 @@ function processPluginHtml(html, pluginId) {
...
+    var attribute = locItem.attribute;
+    var source, target;
+    if (attribute) {
+        source = attribute + '="' + locItem.source + '"';
+        target = attribute + '="' + locItem.target + '"';
+    } else {
+        source = '>' + locItem.source + '<';
+        target = '>' + locItem.target + '<';
+    }
+    var newHtml = html.replace(source, target);
```
MSLaguana: These replacements will only replace the first instance. Do we have any localized strings which appear multiple times?

tools/i18n/files.js
```diff
@@ -0,0 +1,75 @@
...
+
+    directories.forEach(function (dir) {
+        result = result.concat(findFiles(dir, ext));
+    });
+
+    return result;
+}
+
+function isDirectory(path) {
+    return existsSync(path) && fs.statSync(path).isDirectory();
```
MSLaguana: Why `existsSync` and then `statSync`? Especially since `existsSync` just calls `statSync` in a try/catch

tools/i18n/update.js
```diff
@@ -0,0 +1,262 @@
...
+            if (!locId) {
+                locId = applyId(element)
+            }
+            strings.push({type: 'attribute', id: translatedAttribute + '-' + locId, text: attValue});
+        }
+    });
+}
+
+function getId() {
+    return 'xxxxxxxx'.replace(/[xy]/g, function (c) {
```
MSLaguana: Why is this regex `/[xy]/` when the input consists entirely of `x`?

tools/i18n/update.js
```diff
@@ -0,0 +1,262 @@
...
+}
+
+function applyId(element) {
+    var id = getId();
+    element.attribs[LOC_ID_ATTRIB] = id;
+    return id;
+}
+
+// Recreate 'xlf' files.
+// Recreate 'html.<lang>.json' files.
```
MSLaguana: What are these comments referring to?

MSLaguana:
Right now I think I prefer an alternate approach to applying the translations; rather than modifying the html files dynamically as they are served, I think a more resilient approach would be to produce localized versions at build time (using the kind of scripts that you've written here, and benefiting from proper html/xml parsers) and then just selecting a file to serve at runtime. To me that sounds less likely to fail, and also simpler for other people to adopt in their own plugins, if they wish to provide a localized simulation.

src/plugins/exec/sim-host-dialogs.html
TimBarham: It was more that it made the string to be translated needlessly complicated (full of unnecessary `\n`'s and spaces).

src/server/server.js
TimBarham: Oh yeah, good point (I rearranged this at the last minute and forgot to update that).

src/server/server.js
```diff
@@ -394,4 +404,58 @@ function processPluginHtml(html, pluginId) {
...
+
+        locItems.forEach(function (locItem) {
+            html = replaceLocalizationItem(html, locItem);
+        });
+    }
+    return html;
+}
+
+function replaceLocalizationItem(html, locItem) {
+    // To try and be as targeted as possible, we replace attribute by looking for 'attribute="value"', and
```
TimBarham: Yeah, my thoughts on this were as follows:
* My main focus here was getting the translation cycle in place (around the xliff files) ASAP so the loc team could begin their work, rather than necessarily getting the replacement process perfect.
* It works with what we have now.
* Any HTML files that contain strings to be translated are actually rewritten by the `update` process (to insert the loc id attribute), which will clean up white space and quotes.
* If I was going to stick with this approach, I would use a regex here, and/or add some HTML linting to our test process (though the linting that wouldn't help UI defined in plugins).
* If I move to the "generated localized files in the update process" approach, this problem goes away.

Bottom line is my preference would be to leave this as is for now, and iterate on it once we have kicked off the translation process.

src/server/server.js
TimBarham: If they appear multiple times, there will be multiple entries.

tools/i18n/files.js
TimBarham: If I just called `statSync()` I'd have to wrap it in a try/catch anyway, and `existsSync()` does that for me 😄.

tools/i18n/update.js
TimBarham: Oops, copy and paste error. I've replaced it with a less cool but faster version.

tools/i18n/update.js
TimBarham: Steps that I needed to do, and did, but forgot to remove the comments 😄.

TimBarham:
I prefer that approach too and definitely plan on switching to that. I'd just like to get this in to get the ball rolling with the loc team, then make that tweak.

PR #155, opened 2016-09-27
Creates localized HTML files instead of JSON files.

TimBarham:
Rather than doing a replace at runtime, this creates localized versions of each HTML file on update, and serves up the appropriate version of the file at runtime.

This actually ended up being a little bit tricky, so the change is greater in scope than I had originally thought.


src/server/server.js
```diff
@@ -404,59 +408,11 @@ function processPluginHtml(htmlFile, pluginId, lang) {
...
-}
-
-function getI18nJsonFile(htmlFile, lang) {
-    var targetJsonFile = htmlFile.split(path.sep);
-    var pos = targetJsonFile.indexOf('src');
-    targetJsonFile.splice(pos + 1, 0, 'i18n', lang);
-    return targetJsonFile.join(path.sep) + '.i18n.json';
+function getLocalizedHtmlFile(htmlFile, lang) {
+    var localizedHtmlFile = htmlFile.split(path.sep);
+    var pos = localizedHtmlFile.indexOf('src');
```
MSLaguana: How are you finding the localized files? Looks like you are mapping `src/plugins/cordova-plugin-geolocation/sim-host-panels.html` to `src/i18n/<lang>/plugins/cordova-plugin-geolocation/sim-host-panels.html` is that right?

My first thought was this would make it a little harder to work with 3rd party plugins providing their own simulation, and it would be easier to instead map to `src/plugins/cordova-plugin-geolocation/<lang>/sim-host-panels.html` but for a first release I think what you have is fine.

tools/i18n/update.js
```diff
@@ -49,62 +49,100 @@ htmlFiles.forEach(function (htmlFile) {
...
+                    htmlparser2TreeAdapter.insertBefore(parentElement, node, nextSibling);
+                } else {
+                    htmlparser2TreeAdapter.appendChild(parentElement, node);
+                }
+            });
+        }
+    });
+}
+
+function writeFileSync(file, contents) {
```
MSLaguana: This looks like a candidate for the `files.js` helper file.

tools/i18n/update.js
```diff
@@ -146,64 +194,113 @@ function applyTranslations(items, idsRequiringTranslation, targetLanguage) {
...
 
 function serializeNode(node) {
+    // This extracts a node to a document fragment then serialized the fragment. We need to remember parent, prev and
+    // next so we can restore them when we're done (appendChild() doesn't remove the node from its previous parent's
+    // children, so we don't have to patch that).
+    var parent = node.parent;
+    var prev = node.prev;
+    var next = node.next;
+
+    var parentNode = node.parent;
```
MSLaguana: Unused

MSLaguana:
Looks good to me otherwise

src/server/server.js
TimBarham: Yeah, my thinking was I preferred the localized files all together in a mirrored folder structure, rather than scattered through-out. I've added logic to look under the folder that contains the file if we don't find it in the other location.

This also made me think of something - I now add a "Do not edit this file. It is machine generated" comment to the start of each generated HTML file.

tools/i18n/update.js
TimBarham: Hah, yeah, I guess it does :)

tools/i18n/update.js
TimBarham: Removed.

PR #156, opened 2016-09-27
Generate a single xliff file for each language.

TimBarham:
Rather than one per source file per language. At the request of our loc team.


tools/i18n/update.js
```diff
@@ -69,22 +72,30 @@ htmlFiles.forEach(function (htmlFile) {
         return previous;
     }, {});
 
-    updateLanguageXlfJson(sourceHtmlRelativePath, strings).then(function (result) {
+    return updateLanguageXlfJson(sourceHtmlRelativePath, strings).then(function (result) {
```
albertinadx: I would prefer if we can put `then`, `fail` etc callbacks in a new line, to improve readability. That was an implicit convention that we used to follow.
```
return updateLanguageXlfJson(/* ... */)
    .then(function () { 
        // implementation here
    });

```

albertinad: I would prefer if we can put then, fail etc callbacks in a new line, to improve readability. That was an implicit convention that we used to follow.
```
return updateLanguageXlfJson(/* ... */)
    .then(function () { 
        // implementation here
    });
```

albertinad:
LGTM!

tools/i18n/update.js
TimBarham: Sure, will fix. Though I hate the extra indentation that results. One day we'll move this to TypeScript and use await/async and avoid all this awkward Promise stuff 😄.

TimBarham:
PR auto-closed because I deleted the base branch. Made the commit direct to master.

PR #158, opened 2016-09-28
Update Chinese language names and sort XLIFF files by source file order

TimBarham:
The order we process languages can change when we get machine generated translations - depends how quickly they come back. The result is that the order in which we construct the XLIFF files can change. This change sort them first so the order remains consistent.

Also, per loc team request, update Chinese language names.


TimBarham:
Trivial change so I'm just going to merge, as I want to subsequently merge the updated XLIFF and HTML files that result from this change.

PR #165, opened 2016-10-05
Munge ids within custom elements to ensure they're unique.

TimBarham:
This ensures, for example, that labels and their target elements are correctly matched, which will be important for accessibility (keyboard access once we add access keys, and screen readers).

### DIFF
--- a/src/sim-host/ui/custom-elements.js
+++ b/src/sim-host/ui/custom-elements.js
@@ -3,6 +3,8 @@
 var dialog = require('dialog'),
     utils = require('utils');
 
+var uniqueIdSuffix = 0;
+
 function initialize(changePanelVisibilityCallback) {
     registerCustomElement('cordova-panel', {
         proto: {
@@ -21,8 +23,7 @@ function initialize(changePanelVisibilityCallback) {
             }
         },
         initialize: function () {
-            var panel = this;
-            var content = panel.shadowRoot.querySelector('.cordova-content');
+            var content = this.shadowRoot.querySelector('.cordova-content');
             var panelId = this.getAttribute('id');
             var collapseIcon = this.shadowRoot.querySelector('.cordova-collapse-icon');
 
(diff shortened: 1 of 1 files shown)


TimBarham:
When looking at the changes, I'd suggest adding `?w=1` to the end of the URL to ignore whitespace changes.

src/sim-host/ui/custom-elements.js
```diff
@@ -412,6 +480,14 @@ function registerCustomElement(name) {
     });
 }
 
+function getId() {
+    var result = [];
+    for (var i = 0; i < 8; i++) {
+        result.push((Math.random() * 16 | 0).toString(16));
```
MSLaguana: If the idea is to ensure everything gets a unique ID, why not just use an incrementing counter?

MSLaguana: I also believe that IDs must start with a letter and not a number, which means this can generate some invalid identifiers since we use it as a prefix

MSLaguana: Oh hey, in HTML5 ids are super broad and can start with numbers, so feel free to ignore the second point there.

albertinad:
LGTM!

src/sim-host/ui/custom-elements.js
TimBarham: > why not just use an incrementing counter

Er, because that isn't as cool looking? 😄

src/sim-host/ui/custom-elements.js
```diff
@@ -5,331 +5,383 @@ var dialog = require('dialog'),
...
-            var collapsed = collapseIcon.classList.contains('cordova-collapsed');
-
-            if (collapsed) {
-                expandPanel(collapseIcon, content);
-            } else {
-                collapsePanel(collapseIcon, content);
-            }
+        },
+        initialize: function () {
+            var panel = this;
```
TimBarham: Indeed. Fixed.

PR #168, opened 2016-10-12
Scale the compass widget based on current panel width.

TimBarham:
This is important to handle themeing scenarios (where font and panel sizes might be larger, for example). Other fixes:

* Ensure geolocation and compass help text is centered even when it doesn't fill the panel width.
* Center E and W vertically in compass widget.
* Correctly initialize compass direction to "N" rather than "NW".


BretJohnson:
Tim and I chatted some on the phone about this, about related issues with themes/accessibility tenant.  But these changes LGTM.

PR #172, opened 2016-10-16
Improve the appearance of the compass.

TimBarham:
Replaces current compass with a nice pseudo-photo-realistic compass.

Copyright info: The SVG for the compass was created by hand, entirely by me. The design of the compass face was visually based on some free (and copyright-free) [clip art](https://openclipart.org/detail/95047/compass) (released under CC0 license, which states, in part “You can copy, modify, distribute and perform the work, even for commercial purposes, all without asking permission”), but I didn’t reference the original SVG file at all when creating this one.

For comparison, here's the original compass:

image

And here's the new compass:

image

### DIFF
--- a/src/plugins/cordova-plugin-device-orientation/compass-widget.js
+++ b/src/plugins/cordova-plugin-device-orientation/compass-widget.js
@@ -2,22 +2,6 @@
 
 var navigationUtils = require('utils').navHelper();
 
-var COMPASS_NEEDLE = {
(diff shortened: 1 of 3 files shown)


TimBarham:
It was suggested the appearance of the compass was too skeuomorphic, so I've updated it to be less so:

image

This new version has the advantage that it could also change colors to reflect the current theme, once we support themes.

PR #176, opened 2016-10-26
Fix cordova-item delete button to work with themes.

TimBarham:
Old approach used a PNG and a fixed white background. This new approach uses a button (which will be styled with the current theme's button styling) with an SVG that will scale:

image

I think the result is actually nicer than we had before.


MSLaguana:
Looks good to me.

TimBarham:
And here it is with scaled to a theme with a larger font size:

image

PR #182, opened 2016-10-28
Disable panels when dialog showing.

TimBarham:


src/modules/sim-host/dialog.js
```diff
@@ -28,6 +31,15 @@ function showDialog(dialogId, cb) {
 
     currentDialogId = dialogId;
 
+    // Disable the panels if they're not already
+    if (!panelsDisabled) {
+        activeElement = document.activeElement;
+        Array.prototype.forEach.call(getPanels(), function (panel) {
```
MSLaguana: What is the benefit of this over `getPanels().forEach(function (panel) {` ?

TimBarham: `getPanels()` returns a `nodeset`, not an array, so we can't call forEach() (well, recent versions of JS support `forEach()` on a `nodeset`, but we can't rely on that).

PR #183, opened 2016-10-28
Don't translate two letter compass directions.

TimBarham:
(Single letter directions already handled)


albertinad:
LGTM!

TimBarham:
Thanks @albertinad - glad you're still hanging around 😄

PR #185, opened 2016-11-03
Make sure cordova-simulate works well with screen reader

jicongw:
This change resolves a few issues when Windows Narrator reads out cordova-simulate interface.
1. Style tags are read out for a few plugins - change to inline CSS or move to css style sheet
2. Slider, or input type is range, does not have label associated - add "for" attribute to label so that it gets read out
3. Checkbox does not have label associated - add munge id when creating element so that the label gets read out


TimBarham:
Thanks @jicongw!

PR #186, opened 2016-11-03
screen reader reads out device-orientation compass svg text

jicongw:
add aria-hidden properties to text tag so it gets ignored by screen reader


TimBarham:
So you have to add the `aria-hidden` attribute to each `text` element - you can't just add it to the overall `svg` element?

jicongw:
good point. validated and changed accordingly.

TimBarham:
Sweet, thanks.

PR #187, opened 2016-11-03
Keyboard support for persisted exec calls panel.

TimBarham:
Each item can now get the focus, and hitting the `delete` key while it has the focus will delete it. The up and down arrow keys also move up and down the list when an item in the list has the focus.


TimBarham:
Thanks @MSLaguana!

PR #189, opened 2016-11-04
More enhancement on screen reader scenarios

jicongw:
1. For motion plugin - a) make sure text under image is not read out
2. For orientation plugin - a) make sure text under image is not read out b) make sure label for text box is not read twice when focus on the panel
3. For geolocation plugin - a) make sure geo-delay slider value is not read twice b) make sure the map is not read
4. In general, add "panel" to each panel caption when the caption is read out

### DIFF
--- a/src/plugins/cordova-plugin-battery-status/sim-host-panels.html
+++ b/src/plugins/cordova-plugin-battery-status/sim-host-panels.html
@@ -3,11 +3,11 @@
 <cordova-panel id="battery-status" caption="Battery Status" panel-label="Battery Status Panel" data-loc-id="ac1dbd34">
     <cordova-group>
         <cordova-panel-row>
-            <cordova-label read-out="true" for="battery-level" label="Battery level" data-loc-id="e0d4e236"></cordova-label>
+            <cordova-label read-out for="battery-level" label="Battery level" data-loc-id="e0d4e236"></cordova-label>
             <cordova-label id="battery-level-label" label="100%" style="margin-right:5px;margin-left:auto;"></cordova-label>
             <input id="battery-level" type="range" value="100" min="0" max="100">
         </cordova-panel-row>
     </cordova-group>
 
-    <cordova-checkbox id="is-plugged" checked="" read-out="true" data-loc-id="10cfe833">Device plugged</cordova-checkbox>
+    <cordova-checkbox id="is-plugged" checked="" read-out data-loc-id="10cfe833">Device plugged</cordova-checkbox>
 </cordova-panel>
--- a/src/plugins/cordova-plugin-battery-status/sim-host-panels.html
+++ b/src/plugins/cordova-plugin-battery-status/sim-host-panels.html
@@ -3,11 +3,11 @@
 <cordova-panel id="battery-status" caption="Battery Status" panel-label="Battery Status Panel" data-loc-id="ac1dbd34">
     <cordova-group>
         <cordova-panel-row>
-            <cordova-label for="battery-level" label="Battery level" data-loc-id="e0d4e236"></cordova-label>
+            <cordova-label read-out="true" for="battery-level" label="Battery level" data-loc-id="e0d4e236"></cordova-label>
             <cordova-label id="battery-level-label" label="100%" style="margin-right:5px;margin-left:auto;"></cordova-label>
             <input id="battery-level" type="range" value="100" min="0" max="100">
         </cordova-panel-row>
     </cordova-group>
 
-    <cordova-checkbox id="is-plugged" checked="" data-loc-id="10cfe833">Device plugged</cordova-checkbox>
+    <cordova-checkbox id="is-plugged" checked="" read-out="true" data-loc-id="10cfe833">Device plugged</cordova-checkbox>
 </cordova-panel>
--- a/src/plugins/cordova-plugin-battery-status/sim-host-panels.html
+++ b/src/plugins/cordova-plugin-battery-status/sim-host-panels.html
@@ -1,13 +1,13 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
 
-<cordova-panel id="battery-status" caption="Battery Status" panel-label="Battery Status Panel" data-loc-id="ac1dbd34">
+<cordova-panel id="battery-status" caption="Battery Status" spoken-text="Battery Status Panel" data-loc-id="ac1dbd34">
     <cordova-group>
         <cordova-panel-row>
-            <cordova-label read-out for="battery-level" label="Battery level" data-loc-id="e0d4e236"></cordova-label>
+            <cordova-label spoken for="battery-level" label="Battery level" data-loc-id="e0d4e236"></cordova-label>
             <cordova-label id="battery-level-label" label="100%" style="margin-right:5px;margin-left:auto;"></cordova-label>
             <input id="battery-level" type="range" value="100" min="0" max="100">
         </cordova-panel-row>
     </cordova-group>
 
-    <cordova-checkbox id="is-plugged" checked="" read-out data-loc-id="10cfe833">Device plugged</cordova-checkbox>
+    <cordova-checkbox id="is-plugged" checked="" spoken data-loc-id="10cfe833">Device plugged</cordova-checkbox>
 </cordova-panel>
--- a/src/plugins/cordova-plugin-battery-status/sim-host-panels.html
+++ b/src/plugins/cordova-plugin-battery-status/sim-host-panels.html
@@ -1,6 +1,6 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
 
-<cordova-panel id="battery-status" caption="Battery Status" data-loc-id="ac1dbd34">
+<cordova-panel id="battery-status" caption="Battery Status" panel-label="Battery Status Panel" data-loc-id="ac1dbd34">
     <cordova-group>
         <cordova-panel-row>
             <cordova-label for="battery-level" label="Battery level" data-loc-id="e0d4e236"></cordova-label>
--- a/src/plugins/cordova-plugin-camera/sim-host-panels.html
+++ b/src/plugins/cordova-plugin-camera/sim-host-panels.html
@@ -1,6 +1,6 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
 
-<cordova-panel id="camera" caption="Camera" data-loc-id="553fc9a5">
+<cordova-panel id="camera" caption="Camera" panel-label="Camera Panel" data-loc-id="553fc9a5">
     <cordova-group id="camera-options">
         <!--<cordova-radio id="camera-host">Use host system camera</cordova-radio>-->
         <cordova-radio id="camera-prompt" checked="true" data-loc-id="4f087d38">Prompt me for a file</cordova-radio>
--- a/src/plugins/cordova-plugin-camera/sim-host-panels.html
+++ b/src/plugins/cordova-plugin-camera/sim-host-panels.html
@@ -1,6 +1,6 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
 
-<cordova-panel id="camera" caption="Camera" panel-label="Camera Panel" data-loc-id="553fc9a5">
+<cordova-panel id="camera" caption="Camera" spoken-text="Camera Panel" data-loc-id="553fc9a5">
     <cordova-group id="camera-options">
         <!--<cordova-radio id="camera-host">Use host system camera</cordova-radio>-->
         <cordova-radio id="camera-prompt" checked="true" data-loc-id="4f087d38">Prompt me for a file</cordova-radio>
--- a/src/plugins/cordova-plugin-device-motion/sim-host-panels.html
+++ b/src/plugins/cordova-plugin-device-motion/sim-host-panels.html
@@ -1,6 +1,6 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
 <link href="sim-host.css" rel="stylesheet">
-<cordova-panel id="device-motion" caption="Accelerometer" panel-label="Accelerometer Panel" data-loc-id="9a8c8723">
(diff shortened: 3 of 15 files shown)


MSLaguana:
Looks good to me

src/sim-host/ui/sim-host.html
```diff
@@ -16,7 +16,7 @@
          getting cropped in Chrome when wrapping columns -->
     <template id="cordova-panel-template">
         <section class="cordova-panel-inner" tabindex="0">
-            <section class="cordova-header"><span></span><span class="cordova-collapse-icon"><svg focusable="false" viewBox="0 0 14 14">
+            <section class="cordova-header"><div id="panel-label" /><span aria-hidden="true"></span><span class="cordova-collapse-icon"><svg focusable="false" viewBox="0 0 14 14">
```
jicongw: Fixed.

src/sim-host/ui/sim-host.html
```diff
@@ -27,8 +27,8 @@
     </template>
 
     <template id="cordova-dialog-template">
-        <section class="cordova-panel-inner">
-            <section class="cordova-header"><span></span><span class="cordova-close-icon"><svg focusable="false" viewBox="0 0 14 14">
+        <section class="cordova-panel-inner" tabindex="0">
```
TimBarham: With this change, what happens when the dialog displays? We probably want the dialog title and contents read out. Currently there is logic to set the focus to the first focusable element in the dialog rather than to the dialog itself. Does that logic now set the focus to the dialog, or should we change it to do so?

jicongw: Right now it does not focus the dialog after switching back to VS.

src/sim-host/ui/custom-elements.js
```diff
@@ -262,6 +266,10 @@ function initialize(changePanelVisibilityCallback) {
                 // Reverse the order of the checkbox and caption
                 this.shadowRoot.appendChild(this.shadowRoot.querySelector('label'));
             }
+            var shouldReadOut = this.getAttribute('read-out');
+            if (shouldReadOut && shouldReadOut == 'true') {
```
TimBarham: You should implement this as a standard boolean attribute. That means you specify it without a value:

    <cordova-label read-out label="No values saved">

And you should be able to check for it simply using `this.hasAttribute('read-out')`

src/plugins/cordova-plugin-camera/sim-host-panels.html
```diff
@@ -1,6 +1,6 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
 
-<cordova-panel id="camera" caption="Camera" data-loc-id="553fc9a5">
+<cordova-panel id="camera" caption="Camera" panel-label="Camera Panel" data-loc-id="553fc9a5">
```
TimBarham: It seems `panel-label` and `read-label` mean basically the same thing. For consistency we should replace them with a single attribute that is used for both. Perhaps `spoken-text` or something like that. 

If we *do* use `spoken-text`, then we should change `read-out` to something consistent (I think `spoken` and related words is clearer than `read`) - `read-out` could just become `spoken`.

src/sim-host/ui/custom-elements.js
jicongw: Fixed in the next commit.

src/sim-host/ui/sim-host.html
```diff
@@ -16,7 +16,7 @@
          getting cropped in Chrome when wrapping columns -->
     <template id="cordova-panel-template">
         <section class="cordova-panel-inner" tabindex="0">
-            <section class="cordova-header"><span></span><span class="cordova-collapse-icon"><svg focusable="false" viewBox="0 0 14 14">
+            <section class="cordova-header"><span id="panel-label"><span aria-hidden="true"></span></span><span class="cordova-collapse-icon"><svg focusable="false" viewBox="0 0 14 14">
```
jicongw: Fixed in the latest commit.

src/plugins/cordova-plugin-device-motion/sim-host-panels.html
```diff
@@ -1,20 +1,20 @@
...
+        <em id="accelerometer-help" aria-hidden="true" data-loc-id="c562ff87">Click and drag with the mouse to manipulate the device. <br> Hold the Shift key to modify 'alpha'</em>
     </cordova-panel-row>
 
-    <cordova-labeled-value id="accel-x" label="X (m/s²)" value="0"></cordova-labeled-value>
-    <cordova-labeled-value id="accel-y" label="Y (m/s²)" value="0"></cordova-labeled-value>
-    <cordova-labeled-value id="accel-z" label="Z (m/s²)" value="0"></cordova-labeled-value>
-    <cordova-labeled-value id="accel-alpha" label="alpha (deg)" value="0"></cordova-labeled-value>
-    <cordova-labeled-value id="accel-beta" label="beta (deg)" value="0"></cordova-labeled-value>
-    <cordova-labeled-value id="accel-gamma" label="gamma (deg)" value="0"></cordova-labeled-value>
+    <cordova-labeled-value id="accel-x" label="X (m/s²)" read-label="X meters per second squared" value="0"></cordova-labeled-value>
```
jicongw: Fixed in next commit.

src/plugins/cordova-plugin-camera/sim-host-panels.html
jicongw: Fixed in next commit

src/plugins/cordova-plugin-battery-status/sim-host-panels.html
```diff
@@ -1,13 +1,13 @@
...
         <cordova-panel-row>
-            <cordova-label for="battery-level" label="Battery level" data-loc-id="e0d4e236"></cordova-label>
+            <cordova-label spoken for="battery-level" label="Battery level" data-loc-id="e0d4e236"></cordova-label>
             <cordova-label id="battery-level-label" label="100%" style="margin-right:5px;margin-left:auto;"></cordova-label>
             <input id="battery-level" type="range" value="100" min="0" max="100">
         </cordova-panel-row>
     </cordova-group>
 
-    <cordova-checkbox id="is-plugged" checked="" data-loc-id="10cfe833">Device connected to power</cordova-checkbox>
+    <cordova-checkbox id="is-plugged" checked="" spoken data-loc-id="10cfe833">Device connected to power</cordova-checkbox>
```
TimBarham: Why do some checkboxes have the `spoken` attribute and some don't? Why do checkboxes even need this attribute (when would a checkbox ever not need to be spoken)?

src/plugins/cordova-plugin-battery-status/sim-host-panels.html
```diff
@@ -1,13 +1,13 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
 
-<cordova-panel id="battery-status" caption="Battery Status" data-loc-id="ac1dbd34">
+<cordova-panel id="battery-status" caption="Battery Status" spoken-text="Battery Status Panel" data-loc-id="ac1dbd34">
     <cordova-group>
         <cordova-panel-row>
-            <cordova-label for="battery-level" label="Battery level" data-loc-id="e0d4e236"></cordova-label>
+            <cordova-label spoken for="battery-level" label="Battery level" data-loc-id="e0d4e236"></cordova-label>
```
TimBarham: Is the `spoken` attribute needed here because it isn't read with the input element?

src/plugins/cordova-plugin-device-motion/sim-host-panels.html
```diff
@@ -1,20 +1,20 @@
...
-    <cordova-number-entry id="accel-y" label="Y (m/s²)" value="0" step="any"></cordova-number-entry>
-    <cordova-number-entry id="accel-z" label="Z (m/s²)" value="0" step="any"></cordova-number-entry>
-    <cordova-number-entry id="accel-alpha" label="alpha (deg)" value="0" min="0" max="360" step="1"></cordova-number-entry>
-    <cordova-number-entry id="accel-beta" label="beta (deg)" value="0" min="-180" max="180" step="1"></cordova-number-entry>
-    <cordova-number-entry id="accel-gamma" label="gamma (deg)" value="0" min="-90" max="90" step="1"></cordova-number-entry>
+    <cordova-number-entry id="accel-x" label="X (m/s²)" spoken-text="X (meters per second squared)" value="0" step="any"></cordova-number-entry>
+    <cordova-number-entry id="accel-y" label="Y (m/s²)" spoken-text="Y (meters per second squared)" value="0" step="any"></cordova-number-entry>
+    <cordova-number-entry id="accel-z" label="Z (m/s²)" spoken-text="Z (meters per second squared)" value="0" step="any"></cordova-number-entry>
+    <cordova-number-entry id="accel-alpha" label="alpha (deg)" spoken-text="alpha (degree)" value="0" min="0" max="360" step="1"></cordova-number-entry>
```
TimBarham: Should be `degrees`.

src/sim-host/ui/custom-elements.js
```diff
@@ -53,8 +53,9 @@ function initialize(changePanelVisibilityCallback) {
             var panelId = this.getAttribute('id');
             var collapseIcon = this.shadowRoot.querySelector('.cordova-collapse-icon');
 
-            this.shadowRoot.querySelector('.cordova-header span').textContent = this.getAttribute('caption');
-
+            this.shadowRoot.querySelector('.cordova-header .spoken-text span').textContent = this.getAttribute('caption');
+            this.shadowRoot.querySelector('.cordova-header .spoken-text').setAttribute('aria-label', this.getAttribute('spoken-text'));
```
TimBarham: This is a nit, because we always specify `spoken-text`, but really this should be something like:

    this.shadowRoot.querySelector('.cordova-header .spoken-text').setAttribute('aria-label', this.getAttribute('spoken-text') || this.getAttribute('caption'));

src/sim-host/ui/custom-elements.js
```diff
@@ -100,7 +102,9 @@ function initialize(changePanelVisibilityCallback) {
             }
         },
         initialize: function () {
-            this.shadowRoot.querySelector('.cordova-header span').textContent = this.getAttribute('caption');
+            this.shadowRoot.querySelector('.cordova-header .spoken-text span').textContent = this.getAttribute('caption');
+            this.shadowRoot.querySelector('.cordova-header .spoken-text').setAttribute('aria-label', this.getAttribute('spoken-text'));
```
TimBarham: Same comment as for `cordova-panel` - fall back on `caption`.

src/sim-host/ui/custom-elements.js
```diff
@@ -375,12 +388,19 @@ function initialize(changePanelVisibilityCallback) {
...
             this.classList.add('cordova-group');
             this._internalValue = 0;
 
             var input = this.shadowRoot.querySelector('input');
+            var readLabel = this.getAttribute('spoken-text');
+            if (readLabel) {
+                input.setAttribute('aria-label', readLabel);
+            } else {
+                input.setAttribute('aria-label', displayLabel);
+            }
```
TimBarham: You can replace the above few lines with a single line:

    input.setAttribute('aria-label', this.getAttribute('spoken-text') || displayLabel);

src/sim-host/ui/custom-elements.js
```diff
@@ -442,7 +462,16 @@ function initialize(changePanelVisibilityCallback) {
...
-            this.shadowRoot.querySelector('label').textContent = this.getAttribute('label');
+            var label = this.shadowRoot.querySelector('label');
+            label.textContent = this.getAttribute('label');
+            var readLabelDiv = this.shadowRoot.querySelector('.spoken-text');
+            var readLabel = this.getAttribute('spoken-text');
+            if (readLabel) {
+                readLabelDiv.setAttribute('aria-label', readLabel);
+            } else {
+                readLabelDiv.setAttribute('aria-label', label.textContent);
+            }
```
TimBarham: As above - replace this with a single line.

src/sim-host/ui/sim-host.html
```diff
@@ -64,24 +64,24 @@
...
     </template>
 
     <template id="cordova-label-template">
-        <label></label>
+        <label aria-hidden="true"></label>
     </template>
 
     <template id="cordova-labeled-value-template">
-        <label></label>
+        <div class="spoken-text"><label aria-hidden="true"></label></div>
```
TimBarham: Use a `span`, not a `div` (spans are inline elements, like labels, where-as divs are block elements).

jicongw: This change is reverted because it was primarily for Accelerometer panel where we want to support X(m/s^2). Now that they become edit, this change is not necessary anymore.

src/plugins/cordova-plugin-battery-status/sim-host-panels.html
jicongw: This is because it won't read with the full panel focused.

src/plugins/cordova-plugin-device-motion/sim-host-panels.html
jicongw: Fixed in the commit.

src/sim-host/ui/custom-elements.js
jicongw: Fixed in next commit.

src/sim-host/ui/custom-elements.js
jicongw: This is removed.

src/sim-host/ui/custom-elements.js
jicongw: Fixed in the next commit.

src/sim-host/ui/custom-elements.js
```diff
@@ -90,6 +91,7 @@ function initialize(changePanelVisibilityCallback) {
                     if (focusElement) {
                         focusElement.focus();
                     }
+                    this.querySelector('.cordova-panel-inner').focus();
```
jicongw: Fixed in the next commit

src/sim-host/ui/custom-elements.js
jicongw: Fixed in the next commit.

PR #190, opened 2016-11-04
Make Accelerometer panel values editable.

TimBarham:


src/plugins/cordova-plugin-device-motion/device-motion-model.js
```diff
@@ -1,47 +1,176 @@
...
+
+function normalizeFromAcceleration(values) {
+    values.x = constrainAcceleration(values.x);
+    values.y = constrainAcceleration(values.y);
+    values.z = constrainAcceleration(values.z);
 
-    if (values.x) {
-        _axis.x = parseFloat(values.x);
-        updatedValues.x = _axis.x;
+    var sinY = -values.y / G_CONSTANT;
```
MSLaguana: What is the expected behavior if someone enters unexpected inputs, e.g. `y = 2g`, representing the phone being pushed up/down faster than a freefall or some such? By my reading of the current code, since the absolute value of the acceleration is assumed to be a constant `G_CONSTANT`, that will give incorrect answers.

Instead, should we compute something like `accelerationMagnitude = math.sqrt(values.x * values.x + values.y * values.y + values.z * values.z)` and use that as the denominator here? I think we also need to special-case the `values.x = values.y = values.z = 0` case since that doesn't have unique rotations?

TimBarham: Yeah, it's a valid point. Here was the thought process I went through when I made this change:

My goal was to support keyboard accessibility for this panel as simply as possible. But making the acceleration fields editable opens up a can of worms. First problem was it meant a bunch more work (for the reverse math, and keeping things in sync), and second is the point you raise.

I considered making only the rotation fields editable (and appropriately clamped) - since that is effectively the only functionality available today. But that seemed arbitrarily limiting - there are valid scenarios where you would want to test more extreme acceleration values.

So if we make the acceleration fields editable, we have to deal with the issue you raised, and my conclusion on that was as follows:

In our case, rotation really only exists as a way to define acceleration values (we never report rotation to the app). And going from rotation to acceleration, the magnitude of the acceleration vector will always, of course, be g.

But converting from acceleration to rotation, the magnitude of the acceleration vector could be anything. And if it is anything other than g, there are no possible valid rotation values (assuming a non-accelerating device). If we assume an actual acceleration component plus rotation, how much acceleration and in what direction? Bottom line it seemed to me it didn't really matter... If someone is focusing on acceleration values, the rotation probably isn't that important. And anything we do is wrong one way or another, so we might as well go with what is simplest 😄.

But I'm not strongly tied to that conclusion, so if you feel strongly the other way, let me know.

MSLaguana: Why do you say that there is no way to work out rotation from acceleration if it isn't `1g`? I agree that when specifying rotations it is reasonable to assume the device is approximately stationary and thus the acceleration is `1g`, but when specifying acceleration why not compute the magnitude from the given values and work with that? I think the rest of the maths should follow the same, the only special case is when acceleration is 0 in which case it could have any rotation,

That said, it might be a bit confusing to work with, and there are subtleties like `acceleration = 0` means "freefall" rather than "stationary".

TimBarham:
After offline discussion, agreed to go with it as-is.

PR #204, opened 2016-11-21
Fixes a localization issue in Globalization plugin.

changsi-an:
@TimBarham @jicongw

### DIFF
--- a/src/plugins/cordova-plugin-globalization/sim-host-handlers.js
+++ b/src/plugins/cordova-plugin-globalization/sim-host-handlers.js
@@ -296,8 +296,7 @@ function dstOffsetAbs(date) {
 }
 
 function getLocaleName () {
-    var localeList = document.querySelector('#locale-list');
-    return globalizationData.locales[localeList.selectedIndex];
+    return document.querySelector('#locale-list').value;
 }
 
 module.exports = {
--- a/src/plugins/cordova-plugin-globalization/sim-host-panels.html
+++ b/src/plugins/cordova-plugin-globalization/sim-host-panels.html
@@ -1,7 +1,22 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
 
 <cordova-panel id="globalization" caption="Globalization" spoken-text="Globalization Panel" data-loc-id="5ffc4b1e">
-    <cordova-combo id="locale-list" label="Locale" data-loc-id="9e831385"></cordova-combo>
+    <cordova-combo id="locale-list" label="Locale" data-loc-id="9e831385">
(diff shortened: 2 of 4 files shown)


changsi-an:
Revised the commit. So it only reveals the real source authoring that are not auto-generated by the translation tool.

MSLaguana:
I see that `globalization-data.json` is used elsewhere, and in particular the `locales` property is accessed at https://github.com/Microsoft/cordova-simulate/blob/23aeb240048004d4160b2237396beb5713354fea/src/plugins/cordova-plugin-globalization/sim-host-handlers.js#L300

Should `locales` be removed from the json and into the html, since it is a standard and not something that should actually be localized?

changsi-an:
@MSLaguana  Are you saying taking the "value" attribute of the option element directly?

MSLaguana:
Yes that's the kind of thing I meant.
This looks good to me now.

src/plugins/cordova-plugin-globalization/sim-host-handlers.js
```diff
@@ -297,7 +297,8 @@ function dstOffsetAbs(date) {
 
 function getLocaleName () {
     var localeList = document.querySelector('#locale-list');
-    return globalizationData.locales[localeList.selectedIndex];
+    var options = localeList.shadowRoot.querySelectorAll('option');
+    return options[localeList.selectedIndex].getAttribute('value');
```
TimBarham: You should be able to just return `localeList.value`.

src/plugins/cordova-plugin-globalization/sim-host-panels.html
```diff
@@ -1,7 +1,22 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. -->
 
 <cordova-panel id="globalization" caption="Globalization" spoken-text="Globalization Panel" data-loc-id="5ffc4b1e">
-    <cordova-combo id="locale-list" label="Locale" data-loc-id="9e831385"></cordova-combo>
+    <cordova-combo id="locale-list" label="Locale" data-loc-id="9e831385">
+    	<option value="en-US" data-loc-id="0ebb8953">English</option>
```
TimBarham: Generally I'd leave out the `data-loc-id` attributes until the commit with the `update-loc` run, but including them now should be fine.

changsi-an:
PTAL

src/plugins/cordova-plugin-globalization/sim-host-panels.html
changsi-an: Cool. I have to run the tool to test my change, or a localized version will not have the change. I'd not change it back and forth this time. So let's leave it for now.

PR #205, opened 2016-11-23
Exposing ability to change theme data.

MSLaguana:
This does not force the sim-host UI to reload, but the next time that you do it will reflect the updated theming.

### DIFF
--- a/src/server/simulator.js
+++ b/src/server/simulator.js
@@ -51,7 +51,7 @@ function Simulator(opts) {
         updateDevice: this.updateDevice
     };
 
-    this._config.theme = theme.createTheme(simulatorProxy, opts.theme);
+    this.updateTheme(opts.theme);
 
     this._project = new Project(simulatorProxy, opts.platform);
     this._server = new SimulationServer(simulatorProxy, this._project, this.hostRoot);
@@ -201,6 +201,10 @@ Simulator.prototype.stopSimulation = function () {
 Simulator.prototype.updateDevice = function (newDevice) {
     this.config.deviceInfo = device.updateDeviceInfo(newDevice);
 };
+
+Simulator.prototype.updateTheme = function (themeData) {
+    this._config.theme = theme.createTheme(this.hostRoot['sim-host'], themeData);
+}
 
 /**
  * Parse the options provided and create the configuration instance for the current
--- a/src/server/socket.js
+++ b/src/server/socket.js
@@ -124,6 +124,10 @@ SocketServer.prototype.reloadSimHost = function () {
     this._emitTo(SIM_HOST, 'refresh');
 };
 
+SocketServer.prototype.rethemeSimHost = function () {
+    this._emitTo(SIM_HOST, 'retheme');
+}
+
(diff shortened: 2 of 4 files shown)


MSLaguana:
Still no automatic reloading of the sim-host UI, but I've extended the protocol with a `retheme` message to trigger a reload, and a function on the `SocketServer` to send that message.

src/sim-host/protocol/socket.js
```diff
@@ -49,6 +49,17 @@ module.exports.initialize = function (pluginHandlers, services) {
         document.location.reload(true);
     });
 
+    socket.on('retheme', function () {
+        var links = document.head.getElementsByTagName('link');
+        var themeLinks = Array.prototype.filter.call(links, function (elt){
+            return elt.href.indexOf("sim-host-theme.css") >= 0;
+        });
```
TimBarham: You should just be able to do `document.head.querySelectorAll('link[href=sim-host-theme.css]')` and avoid this loop.

TimBarham: Or 'document.head.querySelector()' since you only care about one.

src/sim-host/protocol/socket.js
```diff
@@ -49,6 +49,17 @@ module.exports.initialize = function (pluginHandlers, services) {
...
+        if (themeLinks.length > 0) {
+            // Trigger the script to reload
+            themeLinks[0].href = "sim-host-theme.css";
```
TimBarham: Is simply setting it to the same value enough to trigger a refresh?

MSLaguana: Yes. It's actually not "the same value", since the `themeLinks[0].href` value is a fully qualified path, but empirically this works in chrome at least.

src/sim-host/protocol/socket.js
MSLaguana: Done.

toddm:
:shipit:

src/server/simulator.js
```diff
@@ -51,7 +51,7 @@ function Simulator(opts) {
         updateDevice: this.updateDevice
     };
 
-    this._config.theme = theme.createTheme(simulatorProxy, opts.theme);
+    this.updateTheme(opts.theme);
```
changsi-an: I am not exactly sure why we need to change it to updateTheme. What likely to happen is that we will put theme reloading logic in updateTheme() method later, so it completes the them refreshing mission. My interpretation of updateTheme means that "there is already an existing theme, and we are changing/updating it". From this semantic, using updateTheme() here might be over-assuming things in this context.

src/server/socket.js
```diff
@@ -124,6 +124,10 @@ SocketServer.prototype.reloadSimHost = function () {
     this._emitTo(SIM_HOST, 'refresh');
 };
 
+SocketServer.prototype.rethemeSimHost = function () {
```
changsi-an: What is our plan to get this thing invoked? 

Can the updateTheme() method in simulator.js call it?

PR #206, opened 2016-11-28
Automatically trigger a retheme

MSLaguana:
When updating the theme on the Simulator object, automatically propagate the change to the sim host

### DIFF
--- a/src/server/simulator.js
+++ b/src/server/simulator.js
@@ -51,7 +51,7 @@ function Simulator(opts) {
         updateDevice: this.updateDevice
     };
 
-    this.updateTheme(opts.theme);
+    this._createTheme(opts.theme);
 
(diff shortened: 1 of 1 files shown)


src/server/simulator.js
```diff
@@ -204,6 +204,9 @@ Simulator.prototype.updateDevice = function (newDevice) {
 
 Simulator.prototype.updateTheme = function (themeData) {
     this._config.theme = theme.createTheme(this.hostRoot['sim-host'], themeData);
+    if (this._server && this._server.simSocket) {
+        this._server.simSocket.rethemeSimHost();
```
TimBarham: So I understand that `this._server` won't be set when this is called in the initialization scenario, but I don't like relying on subtleties like that. How about a private `_createTheme()` method that does the create bit, and we call that from the constructor and from here?

PR #209, opened 2016-11-29
Fix additional cases where 'alpha' should not be translated.

TimBarham:


toddm:
Should there be some sort of loc comment so that they don't try to re-loc it later?

TimBarham:
Yes, you're probably right @toddm. However this won't get re-translated unless the English text is changed. The `xlf` files are auto-generated from the English HTML files - to support loc comments I'd have to update that logic. So probably won't do it for now.

PR #216, opened 2016-12-09
Fixing styling of sliders in IE high contrast mode

MSLaguana:


MSLaguana:
This is what it looks like in high-contrast mode with this change:
image

Prior to the change, the background of the slider had the same colour as the foreground of the slider tab, which made it very difficult to see.

TimBarham:
Ah, good find. It would be nice if we could get rid of that border, but I can live with it how it is.

PR #219, opened 2016-12-30
Use CSS to scale compass, so it happens automatically

TimBarham:
When theme is changed on the fly, the compass doesn't scale to the new font size because that calculation was made on page load. Use CSS to scale instead, so it scales automatically with the panel width.


TimBarham:
@MSLaguana, @drewgillies and @changsi-an - please review, thanks!

PR #220, opened 2016-12-30
Use CSS to scale accelerometer device, so it happens automatically

TimBarham:
When theme is changed on the fly, the accelerometer device widget doesn't scale to the new font size because that calculation was made on page load. Use CSS to scale instead, so it scales automatically with the panel width.

### DIFF
--- a/src/plugins/cordova-plugin-device-motion/sim-host.css
+++ b/src/plugins/cordova-plugin-device-motion/sim-host.css
@@ -6,8 +6,10 @@
     margin-right: auto;
 }
 
-#accelerometer-canvas {
-    display: block;
-    margin-left: auto;
-    margin-right: auto;
+#device-motion #accelerometer-canvas {
+    margin: 0 auto;
+    position: relative;
+
+    /* Weird percentage is 200/294, which keeps size the same as it used to be */
+    width: 68.03%;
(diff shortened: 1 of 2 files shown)


TimBarham:
@MSLaguana, @drewgillies and @changsi-an - please review, thanks!

src/plugins/cordova-plugin-device-motion/sim-host.css
```diff
@@ -6,8 +6,10 @@
...
-    display: block;
-    margin-left: auto;
-    margin-right: auto;
-}
\ No newline at end of file
+#device-motion #accelerometer-canvas {
+    margin: 0 auto;
+    position: relative;
+
+    /* Weird percentage is 200/294, which keeps size the same as it used to be */
```
MSLaguana: Where does 294 come from? It looks like the previous code was using `200/320` as the scaling factor

TimBarham: 294 is "100%" when the panel is 320px wide.

MSLaguana: Ah, so there is 26px of padding or something? Ok.

PR #221, opened 2017-01-05
Add support for scrollbar colors to themes.

TimBarham:
Note that this currently only works for IE. With Chrome, if you theme the scrollbars you lose the arrows. You can add them back in manually (via a background image), but that's not something I'm going to do right now.

### DIFF
--- a/src/sim-host/ui/theme.js
+++ b/src/sim-host/ui/theme.js
@@ -85,5 +85,34 @@ module.exports = {
         }
     },
 
-    defaultFontSize: 13
+    defaultFontSize: 13,
+
+    doCustom: function (css, themeObject) {
+        // Support scroll bar colors. Note that this is only supported for IE, because Webkit/Chrome makes it too difficult.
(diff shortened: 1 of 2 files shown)


src/sim-host/ui/theme.js
```diff
@@ -85,5 +85,36 @@ module.exports = {
...
 
-    defaultFontSize: 13
+    defaultFontSize: 13,
+
+    doCustom: function (css, themeObject) {
+        // Support scroll bar colors. Note that this is only supported for IE, because Webkit/Chrome makes it too difficult.
+        var scrollBarColors = themeObject.scrollbar && themeObject.scrollbar[''];
+        if (scrollBarColors) {
+            var ieScrollBarColorCssProperties = [];
+            if (scrollBarColors.background) {
```
MSLaguana: What is this check for? It seems to be duplicated further down, and not used elsewhere.

TimBarham: Um... it was just to... um.. make sure you were paying attention.

PR #222, opened 2017-01-11
make label value focusable so it can be read by screen reader

jicongw:


MSLaguana:
Looks good to me

jicongw:
Thanks @MSLaguana ! Will merge it.

PR #223, opened 2017-02-03
Add 'name' property to GPX playback speed combobox.

TimBarham:
Also, update `cordova-combobox` to handle `name` property, and localization tools to treat name property as translatable.


MSLaguana:
Looks good to me.

PR #224, opened 2017-02-21
Remove 'secure' attribute from proxied cookies

ascorbic:
Secure cookies sent by the proxied server are ignored by the simulator, because /xhr_proxy is not served over https. This PR strips the secure attribute from set-cookie headers.


EzioLi01:
Close this PR since it have no further updates for 5+ years, please reopen this if you still want to work on this.

PR #228, opened 2017-04-11
Fix fragment links in AngularJS apps

### LINKED ISSUE #227: Anchor tags in AngularJS app that use a fragment don't work in cordova-simulate
**Repo steps:**
1. Create a Cordova app from the Mobile Angular UI sample at http://mobileangularui.com/demo/#/
2. Launch it in *cordova-simulate*
3. Click any of the links down the left-hand side ("Scroll", "Toggle", "Tabs" etc).

**Expected:**
The selected page opens.

**Actual:**
Nothing happens.

TimBarham:
Set app root URL to just "/" instead of poniting to the start page. Then, when "/" is requested, return the start page.

This fixes #227

### DIFF
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -101,7 +101,7 @@ SimulationServer.prototype.start = function (platform, opts) {
 
             this._urls = {
                 root: urlRoot,
-                app: urlRoot + parseStartPage(projectRoot),
+                app: urlRoot,
                 simHost: urlRoot + 'simulator/index.html'
             };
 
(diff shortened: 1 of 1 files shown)


TimBarham:
@MSLaguana - Jimmy, I was actually wondering the same thing, and had intended to try the app in an Android emulator to understand how it behaved there, but forgot. Tried that this morning, and the app is completely broken running in an Android emulator (I can't even get to the point of trying this functionality). I've reached out to the guy who originally reported this problem to see if he has an example app that demonstrates the problem in cordova-simulate, but works in an emulator.

TimBarham:
Abandoning this PR. See comment on #227.

Issue #126, opened 2016-08-05
Unhandled Exec Call dialog does not specify call arguments


MSLaguana:
The Unhandled Exec Call dialog mentions the name of the service and the function on the service, but does not mention any arguments that may have been specified.

AleksanderStrochkov:
https://github.com/Microsoft/cordova-simulate/pull/239 - PR with fix

PR #231, opened 2017-04-28
#126. Add arguments of function in output of Unhandled Exec Call dialog

dmitrii-snytkin:


src/i18n/cs/plugins/exec/sim-host-dialogs.html
```diff
@@ -4,7 +4,7 @@
 <cordova-dialog id="exec-dialog" caption="Neošetřené volání metody Exec" data-loc-id="f7283850">
     <cordova-group>
         <p data-loc-id="a1a63f26">Pro následující volání metody Exec není k dispozici obslužná rutina:</p>
-        <p style="font-style:italic"><span id="exec-service" data-loc-id="ba2b4c80">služba</span>.<span id="exec-action" data-loc-id="e4c1ba28">akce</span></p>
+        <p style="font-style:italic"><span id="exec-service" data-loc-id="ba2b4c80">služba</span>.<span id="exec-action" data-loc-id="e4c1ba28">akce</span>(<span id="exec-args" data-loc-id="veyymo5t">arguments</span>)</p>
```
AnBucyk: @TimBarham are these localized files generated? Is there a process for making this change without manually editing these files?

TimBarham: These are generated by a tool. It looks like @dmitrii-snytkin ran the tools, however (based on the changes that were generated).

However, usually (to make code reviews easier, apart from anything else), I don't run the tool until I'm about to do a release (basically: let a bunch of changes accumulate, then when about to do a release, run the tool, then notify the translation team, they do their changes in one go, then do the release).

So, @dmitrii-snytkin, maybe remove the `i18n` changes from the PR.

TimBarham:
So, the change looks good. However, I have one concern: you end up with something that looks like this:

image

Looks great, except the text below says "If you wish to always respond to this call with this result" - to me, "this call" now implies the specified service, action *and* arguments, where-as the persisted value will always be returned for the service and action combination, regardless of arguments. I think we need to make that clear before we merge this change.

@drewgillies - do you have any thoughts on this?

src/i18n/cs/plugins/exec/sim-host-dialogs.html
dmitrii-snytkin: No, I changed files manually). Should I revert changes in both `src/i18n` and `tools/i18n` folders?

TimBarham: > Should I revert changes in both `src/i18n` and `tools/i18n` folders?

Yes please. For future reference: if you ever end up doing a release, `npm run update-loc` runs a tool that does this automatically, and I run it just before I do a release each time.

TimBarham:
@jmatthiesen - could you look at the above discussion and let me know if you have any thoughts? Thanks!

TimBarham:
Ping @jmatthiesen

jmatthiesen:
Riffing on the text a bit, how's this?

>Please enter a return value below to simulate a call to this function and choose if this value represents success or failure. If you’d like this value returned for all future calls to this function, select the ‘persist’ option.

TimBarham:
Works for me. I'll make that change and merge this.

TimBarham:
Oops - accidentally hit the `Close and comment` button 😄

PR #232, opened 2017-04-28
fixes #72 - Double submit of "Choose file" in "cordova-camera-plugin"

### LINKED ISSUE #72: [Suggestion] Double submit of "Choose file"
If you click the "Choose file" button of the camera plugin really fast, it'll sometimes show the open file dialog multiple times in response. This could be an issue particularly in slow machines, if the user doesn't sees a response and clicks again. It'd be nice if the second instance could be automatically be omitted.

paymicro:
When you press a button on it, it is blocked until a dialog box appears with the file selected.

### DIFF
--- a/src/plugins/cordova-plugin-camera/sim-host.js
+++ b/src/plugins/cordova-plugin-camera/sim-host.js
@@ -79,8 +79,13 @@ module.exports = function (messages) {
             // Setup handlers for choosing an image in the panel
             filenameInput.accept = 'image/*';
             document.getElementById('camera-choose-filename').addEventListener('click', function () {
+                if (this.disabled) {
+                    return;
+                }
+                this.disabled = true;
                 telemetry.sendUITelemetry(Object.assign({}, baseProps, { control: 'camera-choose-filename' }));
                 filenameInput.input.click();
+                this.disabled = false;
             });
 
             filenameInput.addEventListener('change', function () {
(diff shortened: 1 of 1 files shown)


TimBarham:
Just to confirm - you were able to repro the problem, and this change prevented it from occurring? If so, you'll want to make the same change for the `camera-dialog-choose-filename` button (see [line 102](https://github.com/paymicro/cordova-simulate/blob/fb9371416fdf230229ee44d920680543b2709c12/src/plugins/cordova-plugin-camera/sim-host.js#L102)).

paymicro:
I was able to trigger up to 3 click events leading to 3 successive file dialogs.
This solution disables the button for 3 seconds, which should be sufficient for most cases.
It's also seems to be not possible to know if the dialog box was cancelled by the user so we can't rely on "onchanged" event.

paymicro:
@TimBarham please take a look.

TimBarham:
I never really liked this solution, which is why I didn't merge it at the time (and we've never had anyone actually report the problem AFAIK). However, if we going to do it, it should be far less than 3 seconds time out. 500ms would be more than sufficient, I'd think.

ruslan-bikkinin:
Thanks for responding @TimBarham! Actually, I couldn't really reproduce this problem by myself now, a year after PR has been created, but I synced with @paymicro and he told me that he got permanent reproduction by the time he had submitted PR. Any way, I checked changes using different browsers and operating systems and it worked fine, so that's why I approved this PR. Also, I spoke with @iTOYS (he is the maintainer of this repo, too) and he agreed with you regarding the fact that no one really got complained about problem being discussed and he couldn't reproduce it as well, so I believe we can close it for now and get back here if needed.
@paymicro if you have strong disagreements with us, please let us know.

paymicro:
I think that it is not necessary to apply this fix. Issue is extremely rare and not critical. It adds functionality that almost no one needs. Also, no one could reproduce it, except for me...

ruslan-bikkinin:
Okay then, I am closing this for now.

Issue #217, opened 2016-12-19
Deleting a file that is being watched crashes cordova-simulate


MSLaguana:
When a file modification event comes in, cordova-simulate tries to `statSync` it which throws if the file doesn't exist. It'd be much better if we could instead look for the file to be re-created and resume watching then. Currently an ionic build will clean out the www folder and re-create it, which breaks cordova-simulate.

tomingoglia:
This problem manifests itself in VS 2017 when trying to use Ionic 2 with Live Reload. For anyone following this issue and looking for a workaround, there is a option called **forceprepare** that can be used when **livereload** is set to true. If this is set to true, **livereload** will not crash cordova-simulate. 

A simple workaround is to modify the _simulate.js_ file located in the path: _C:\ProgramData\Microsoft\VisualStudio\MDA\975a939a\taco-toolset-6.3.1_

Add the following code to the default configuration attributes in `var = opts {...}`:
`forceprepare: (args["forceprepare"] || "").toLowerCase() !== "false"`

This will default the value to true rather than false if the value is not specified. I'm not sure what effect this will have on other cordova projects, but this fixes the problem with Ionic 2 projects. Based on reading about **forceprepare** and **cordova prepare**, I doubt it will introduce other issues.

Loving VS Code, VS 2017 RC, and the new cordova-simulate features...great work!

PR #233, opened 2017-04-28
#217. Add exception handler for statSync.

dmitrii-snytkin:


dmitrii-snytkin:
Fix for bug https://github.com/Microsoft/cordova-simulate/issues/217

TimBarham:
LGTM. Although I probably wouldn't bother with the warning, since the only reason `statSync` should throw is if the file/directory no longer exists, and that isn't really an error/warning situation.

TimBarham:
Actually, another thought I had here: when a watched file is deleted, we don't want to just bail. It should probably trigger a reload, since it is a legitimate change to the app. So I think we should probably not return here. But we will likely need to add special handling for deleted files later on.

dmitrii-snytkin:
I think, that deleting of www directory is not a legitimate change, because it brokes a work of application. And to skip of deleting www directory is better, than to react on this change. Anyway, if www directory appears, `handleWatcherEvent` will called and app will reload correctly.

TimBarham:
But the bug isn't about deleting the `www` folder - it's about deleting a watched file in the `www` folder, which is a legit scenario.

dmitrii-snytkin:
yes, I'm wrong.

dmitrii-snytkin:
Will @TimBarham approve PR?)

TimBarham:
It looks like we still need to handle this later on. As this change currently stands, when a file is deleted, we hit a warning [here](https://github.com/Microsoft/cordova-simulate/blob/statSync-exception-handler/src/server/live-reload/live-reload.js#L83) and nothing else happens. I think instead deleting a file should trigger the following:

* In `live-reload.js`, delete the corresponding file in the platform folder.
* In `src/app-host/live-reload-client.js`, a deleted file should trigger reloading the page.

TimBarham:
Ok, I've made the change so I can merge this and do a release.

PR #234, opened 2017-05-02
fixes #47 - Added support host system camera and default camera image

### LINKED ISSUE #47: Support host system camera and default camera image
The camera plugin options for using the host system camera and a built-in camera image have been removed, since they were not currently supported.

These functionalities should be implemented and the UI options to select them re-enabled.

paymicro:
- added support host system camera and default camera image


paymicro:
@TimBarham please take a look.

PR #235, opened 2017-05-03
Fixes #77 - radio-buttons behavior in Chrome browser.

### LINKED ISSUE #77: cordova-radio elements don't work in the latest Chrome
After a recent Chrome update (can't pinpoint which), `<cordova-radio>` elements are no longer mutually exclusive with each other. Clicking a different radio button in the group will select the new one without unselecting the previous one.

It still works fine in other browsers.

paymicro:
Manually unchecking the radio buttons, if they are in cordova-group.


TimBarham:
Change looks good, except the check for **Chrome** needs some tweaking, since it will also detect **Microsoft Edge**.

paymicro:
@TimBarham I fixed the browser check. Please take a look.

ruslan-bikkinin:
I am closing this PR in favor of https://github.com/Microsoft/cordova-simulate/pull/268. Generally, approach provided here could break advanced components because it takes into account only the first upper level of hierarchy of components. #268 fixes radiobuttons for cordova-plugin-camera only. If there will be more broken radiobutton behaviour found in other plugins, we could come back here and propose more global solution.

PR #236, opened 2017-05-03
Fixed bug 385093: Accessibility: MAS42A: HC: Text on "Fire event,Choose file,Go button and Zoom in and Zoom out"  buttons in cordova simulation window is not according to the HighContrast Mode

AleksanderStrochkov:
In case of using Windows high contrast theme Cordova buttons have the same color scheme as for standard Windows buttons. Please take a look at following examples: 
1. Windows black high contrast theme:
Windows buttons
image
and Cordova buttons
image
2. Windows black high contrast theme:
Windows buttons
image
and Cordova buttons
image

### DIFF
--- a/src/sim-host/ui/sim-host.css
+++ b/src/sim-host/ui/sim-host.css
@@ -264,6 +264,21 @@ body /deep/ input[type=range]::-webkit-slider-thumb {
     -webkit-appearance: none;
 }
 
+/* Use correct colors for buttons in IE in high-contrast mode */
+@media screen and (-ms-high-contrast: active) {
+    button {
+        color: buttonText;
+        border-color: buttonText;
+        background-color: buttonFace;
+    }
+
+    button:hover, button:focus {
+        color: highlightText;
+        border-color: buttonText;
+        background-color: highlight;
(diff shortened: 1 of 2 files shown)


TimBarham:
Thanks @AleksanderStrochkov - sorry for the delay in me looking at this. I have a couple of comments:
* This change applies to hover, but I believe the original bug was about when the button has the focus - both are certainly important scenarios.
* I think a better fix would be to work within the `cordova-simulate` theming support. VS will provide a color scheme to use when in high contrast mode, and that is defined on the VS side. It may be we can tweak the colors specified by VS to get the correct result here. While we can fall back on an approach like the one you have used here if we can't get it to work right using that approach, I'd prefer you investigate fixing this from the VS side first. Also, if we do end up going with the approach you have here, it should live in `sim-host.css` rather than `sim-host.html`.

TimBarham:
Ok, another issue I discovered is that this only works for `High Contrast Black` and `High Contrast White` themes, not other themes like `High Contrast #1` and `High Contrast #2`. I have a solution that should work in all cases.

PR #238, opened 2017-05-12
Fix bug 433383: An exception is thrown when the header tag is added

AleksanderStrochkov:
Inject script to 'head' tag only.
Previously, there was an injection to all tags that begin from 'head' (etc. 'header', 'headblablabla').

### DIFF
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -252,7 +252,7 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
             send(request, filePath, {
                 transform: function (stream) {
                     return stream
-                        .pipe(replaceStream(/(<\s*head[\s]*>)/, '$1' + scriptTags))
+                        .pipe(replaceStream(/(<\s*head\b[^>]*>)/, '$1' + scriptTags))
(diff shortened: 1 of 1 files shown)


src/server/server.js
```diff
@@ -252,7 +252,7 @@ SimulationServer.prototype._streamAppHostHtml = function (request, response) {
             send(request, filePath, {
                 transform: function (stream) {
                     return stream
-                        .pipe(replaceStream(/(<\s*head[^>]*>)/, '$1' + scriptTags))
+                        .pipe(replaceStream(/(<\s*head[\s]*>)/, '$1' + scriptTags))
```
TimBarham: This will only match a `<head>` tag that contains "head" and white-space, but nothing else. So if the head tag contains attributes, for example, this will fail.

One possibility would be to take the original regex and include a word boundary after "head": `/(<\s*head\b[^>]*>)/`

TimBarham:
Ping @AleksanderStrochkov

TimBarham:
I want to merge this today so we can do a release, so I made the change myself.

Issue #245, opened 2017-06-05
deviceready has not fired after 5 seconds. chromium


dotnetCarpenter:
No matter what I do, I can never get `simulate` to work, so in the end I created a new project to see if it was my code or if its an issue with **cordova-simulate**.

### Steps to reproduce
1. `cordova create test com.test test`
2. `cd test`
3. `cordova platforms add browser`
4. `simulate browser --target=chromium`

Error:
```
cordova.js:1051 deviceready has not fired after 5 seconds.
cordova.js:1044 Channel not fired: onNativeReady
cordova.js:1044 Channel not fired: onCordovaReady
cordova.js:1044 Channel not fired: onCordovaSimulateReady
```

OS: Ubuntu 16.04.2 LTS
cordova: 6.5.0
simulate: 0.3.12
node: v7.10.0
npm: 4.6.1

nodefish:
I have this issue as well. I noticed the following: cordova-simulate is supposed to open 2 browser windows: 1 with the app, and 1 with the cordova-simulate menu (lets you choose the device, etc).

Only the 1st window (app) was opening, and I was seeing the same error messages as you in the Chrome inspector. Upon closing this window, only THEN did the 2nd window open.

I realize the 2nd window (cordova-simulate menu) is what fires the deviceready event.

Workaround: close the 1st window (app) after it errors out (deviceready has not fired...). The 2nd window should now open. Keep it open, open a new tab and navigate to http://localhost:whatever-port/index.html.

This is what worked for me. I'm on Windows 10 / Chrome.

dotnetCarpenter:
According to @purplecabbage, this issue should be fixed in the upstream [cordova-serve](https://github.com/apache/cordova-serve) repo, which has no gihub issue list.

Ref: https://github.com/apache/cordova-browser/pull/31#issuecomment-309171231

PR #246, opened 2017-06-29
Fixed bug with not fired onDeviceReady event

paymicro:
Fixed bug with not fired onDeviceReady event, due to a broken webSocket connection in debug mode.

At the time of debugging, a loss of the WebSocket connection can occur because of this, the onCordovaSimulateReady and onDeviceReady events are not triggered.


TimBarham:
Is this a fix for #245? If so, it seems the problem there is that the simulation host window never opens. Since the simulation host is what triggers start (once it is initialized), that would be why start never gets fired. So I wonder if the problem really needs to be fixed elsewhere - figuring out why the simulation host window doesn't open.

busykai:
It does seem to me that if `sim-host` never tells `app-host` to start, then it should hang (which it correctly does).

Simply firing `deviceReady` on `reconnect` is not the right thing to do. The bootstrap protocol between `sim-host` and `app-host` should resume on `reconnect`. `app-host` cannot make unilateral decision that it's ready.

paymicro:
@TimBarham this is another bug. It is not due to the fact that in the process of app-host debugging the connection to the server through the sockets is lost.
This fix allows after the reconnect, register on the server and run the onCordovaSimulateReady event.

paymicro:
@busykai In this case, sim-host and app-host are executed correctly. 
A bug may occur if you start debugging app-host between the events onNativeReady and onCordovaSimulateReady.

TimBarham:
Sorry for the delay getting back... my comment still applies - `app-host` should never fire `deviceReady` unless and until it receives `start` from the `sim-host`.

paymicro:
@TimBarham I updated PR. `app-host` not fire `deviceReady`. 
After autoreconnection, he registered on the server. And sends plugin list to `sim-host`, if `deviceReady` not fired.

EzioLi01:
Close this PR since no further updates for 5+ years, please reopen this if you still want to work on this.

PR #265, opened 2018-05-21
Fix rendering in Chrome

### LINKED ISSUE #259: Plugin Simulation window rendering broken in Chrome browser
When opening the Plugin Simulation window in Chrome 64-bit version 64.0.3282.167 for Windows, it looks like this (see image).

image

This deprecation warning occurs in the browser console:

> [Deprecation] /deep/ combinator is no longer supported in CSS dynamic profile.It is now effectively no-op, acting as if it were a descendant combinator. /deep/ combinator will be removed, and will be invalid at M65. You should remove it. See https://www.chromestatus.com/features/4964279606312960 for more details.

iTOYS:
Fixes #259 
screen shot 2018-05-21 at 3 14 40 pm


.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
+{
```
ruslan-bikkinin: @iTOYS how do you think, should we add this line to gitignore?

iTOYS: I guess this is helpful template for contributors.

ruslan-bikkinin: Yes, agreed, but we could use it as well.

TimBarham:
Thanks @iTOYS! I have a couple of questions:

1. Have you tested this in other browsers to ensure nothing is broken?
2. Have you tested this with theming to ensure that still works as expected (Visual Studio for example makes heavy use of `cordova-simulate` theming support)? To test this, you'd either have to hack VS to use your local `cordova-simulate` repo, or let VS build a theme file and pass that to `cordova-simulate` from the command line.

iTOYS:
Hi @TimBarham 
1. Yes, I've been checked Safari and Chrome on Mac and Edge, Chrome, Opera and Firefox on Windows.
2.I'm not sure about theming in VSCode, but I've checked `vscode-cordova-tools` which use `cordova-simulate` under the hood, with this changes.
screen shot 2018-05-22 at 11 06 35 am

Could you please give me a bit more info about using `cordova-simulate` inside VSCode and how I can test it?

Thanks!

TimBarham:
Hey @iTOYS - it's regular VS (not VS Code) that uses `cordova-simulate` themes. Easiest way to test it is probably to use a VS generated theme file from the command line. I've attached one you can use. To use this theme, unzip it, then launch `cordova-simulate` form the  using:
```
simulate <platform> --target=<browser> --theme="<path to theme>\4a40218e3d009a31f4fae94389e2fae8.json"
```
[4a40218e3d009a31f4fae94389e2fae8.zip](https://github.com/Microsoft/cordova-simulate/files/2029245/4a40218e3d009a31f4fae94389e2fae8.zip)

TimBarham:
I did a bit of quick testing. IE looks fine (with and without a custom theme). Chrome (Windows, 66.0.3359.181) is better than it was, but still broken for me (no panel borders):
image
And theme support is also pretty broken:
image
It should look basically like - taken from IE:
image

zeprone:
image

this is what I get from the fresh build on branch fix/chrome-rendering with my macbook on Chrome 66.0.33...

iTOYS:
@TimBarham @zeprone 
There was cache issue, added fix for this!
Checked with VS17, all works fine.

Thanks!

zeprone:
much better ;)

TimBarham:
Thanks for the continued work here, @iTOYS - that fixed it!

With some more testing, I did find a couple of issues with the **Persisted Exec Calls** panel (_src/plugins/exec_).

The first relates to borders around items in this panel. When there is a single item in the list, there should be no borders around it:

image

When there are multiple items, there should only be a border between items:

image

(screen shots above taken from Edge without the PR applied)

With this PR applied, we get the following results in Chrome (left and right borders are suppressed, but top and bottom borders are not):

image

And for multiple items (in addition to the above, we have a double thickness inner border):

image

In Edge and Firefox we don't get any border at all:

image

Also, in Edge and Firefox, the "remove" button is visible when the mouse is anywhere over the page. It should only be visible for the item your are currently hovering over:

image

iTOYS:
@TimBarham 
Could you please approve this PR if you don't have more concerns?

Thanks!

TimBarham:
@iTOYS - yep, I'll take a look at it tonight.

TimBarham:
Hey @iTOYS - thanks, that looks perfect!

However... Apologies... but while testing this I noticed another issue - comboboxes aren't styled correctly. Using that same theme file, here's what they should look like:

### Firefox

_Standard_
image

_Rollover/Focus_
image

### IE

_Standard_
image

_Rollover/Focus_
image

Etc.

With this PR applied, I'm not seeing any styling at all, really, so for example on IE it looks like this:

image

Similarly Firefox, Chrome and Edge I'm seeing default styling rather than the theme style applied. Again, sorry I didn't notice this earlier!

iTOYS:
@TimBarham 
Fixed, after/before screenshots:
image

Thanks a lot for you review!

PR #280, opened 2018-12-26
Fix internal URL link on paragraph #133

kant:


ruslan-bikkinin:
Thanks for your contribution @kant, LGTM!

PR #303, opened 2019-08-09
Add command to generate unique UUIDs for simulated devices with UUID4 id format, styling improvements

mfish33:
For the application I am working on I need unique device identifiers for purposes of registration on a remote server. I added a post-install script that generates random UUIDs for all devices except android because of their non-standerd UUID system.


mfish33:
Can I get some insight into why the build is failing? Right now it says I do not have access to the logs.

SounD120:
Hi @mfish33 . Thanks you for your contribution! Build fails due to ESLint error: 
```
./src/devices/make-uuids.js
  1:1  error  Parsing error: The keyword 'const' is reserved
```
I think replacing `const` keyword by `var` should resolve the problem.

mfish33:
I changed it to var but the build is still failing. Can you please let me know why?

SounD120:
@mfish33 There is another ESLint issue
```
./src/devices/make-uuids.js
  8:42  error  Parsing error: Unexpected token >
```
You can run `gulp` command in the `cordova-simulate` folder. This command will run all required checks locally.

mfish33:
I tried running Gulp locally but I keep getting this error:
Message:
    Failed with 2 errors
Details:
    domainEmitter: [object Object]
    domain: [object Object]
    domainThrown: false
But only with my make

SounD120:
@mfish33 Could you please provide the expected results of these changes? Not sure if this change is safe. What tests were carried out?

mfish33:
I fixed the issues stopping the script from properly executing. I just tested the current build in a test app and it works as expected. After the npm install of Cordova simulate random UUID's are created for each device to allow for fully unique devices. However, I would like some insights to why the build is failing since gulp passes before and after the post-install script runs

SounD120:
@mfish33 In CI pipeline there is a check that there are no changes in the code after `npm install` command. For now it is complaining that UUIDs are different from UUIDs that are commited in Git because of postinstall command. Also, I think that this package can be installed as a dependency to other packages, so maybe it could be better to add a manual command that will update these UUIDs in the file instead of updating them on postinstall.

mfish33:
I changed the script to run when you provide simulate with the command --generateids. While this works as intended there is a warning in the cmd line of 
`SIM WARNING: Could not reload the modified file because fs.watch() didn't specify which file was changed`
even when I do not include the flag. I'm pretty sure this has to do with Chokidar but I'm not sure.

src/devices/make-uuids.js
```diff
@@ -0,0 +1,16 @@
+var fs = require('fs');
+var uuidv4 = require('uuid/v4');
+
+var deviceList = require('./devices.json');
+var UUIDPlatformBlacklist = ['android']
```
SounD120:
```suggestion
var UUIDPlatformBlacklist = ['android'];
```

src/simulate.js
```diff
@@ -9,6 +9,9 @@ var launchBrowser = function (target, url) {
 };
 
 var simulate = function (opts) {
+    if(opts.generateids) {
+        require('./devices/make-uuids')
```
SounD120:
```suggestion
        require('./devices/make-uuids');
```

README.md
```diff
@@ -54,6 +54,7 @@ Where `opts` is an object with the following properties (all optional):
 * **touchevents** - A boolean. Set to `false` to disable the simulation of touch events in App-Host. Defaults to `true`.
 * **simulationpath** - the directory where temporary simulation files are hosted. Defaults to `projectRoot/simulate`.
 * **simhosttitle** - specifies the title of the simulation window. Defaults to `Plugin Simulation`.
+*  **generateids** - A boolean that generates unique ids for simulated devices at startup. Defaults to `false`.
```
SounD120:
```suggestion
* **generateids** - A boolean that generates unique ids for simulated devices at startup. Defaults to `false`.
```

src/devices/make-uuids.js
```diff
@@ -0,0 +1,17 @@
...
+Object.keys(deviceList).filter(function(platform) {
+    return !UUIDPlatformBlacklist.includes(platform);
+}).map(function(platform) {
+    deviceList[platform] = deviceList[platform].map(function(device) {
+        return Object.assign(device, {uuid:uuidv4()});
+    });
+});
+
+/*global __dirname */
+fs.writeFileSync(path.join(__dirname, '/devices.json'), JSON.stringify(deviceList, null, 2));
```
SounD120:
```suggestion
fs.writeFileSync(path.join(__dirname, 'devices.json'), JSON.stringify(deviceList, null, 2));
```

src/devices/make-uuids.js
```diff
@@ -0,0 +1,17 @@
...
+var uuidv4 = require('uuid/v4');
+
+var deviceList = require('./devices.json');
+var UUIDPlatformBlacklist = ['android'];
+
+Object.keys(deviceList).filter(function(platform) {
+    return !UUIDPlatformBlacklist.includes(platform);
+}).map(function(platform) {
+    deviceList[platform] = deviceList[platform].map(function(device) {
+        return Object.assign(device, {uuid:uuidv4()});
```
RedMickey:
```suggestion
        return Object.assign(device, {uuid: uuidv4()});
```

PR #308, opened 2020-01-18
Allow the use for custom middleware to allow for a wider range of plugin support

mfish33:
For the application I am working on we have multiple plugins that can not be supported under the current cordova-simulate infrastructure. Internally we have made patches to the files to support the changes in this PR. Below is a flowchart of the infrastructure that we implemented. The basic workflow is that the app host handlers or sim host handlers can make http requests that can be intercepted and redirected to the desired plugin. The reason we went with http requests is that one of our plugins is a proxy server that redirects incoming requests to locally cached files.

Capture

A future PR could include a default middleware file that could support the effortless writing of custom plugins in nodejs.


mfish33:
I closed the Cordova-Serve PR because it was pointed out that my changes could just be implemented here.

SounD120:
Hi @mfish33 . Thanks a lot for your contributions! You mentioned possible default middleware. Could you please give an example of such middleware?

src/server/server.js
```diff
@@ -90,6 +101,10 @@ SimulationServer.prototype.start = function (platform, opts) {
         noServerInfo: true
     };
 
+    if(this._config.middleware) {
```
RedMickey:
```suggestion
    if (this._config.middleware) {
```

bin/simulate
```diff
@@ -93,6 +95,7 @@ function getUsageString() {
         '[--target TARGET]',
         '[--lang LANG]',
         '[--theme THEME]',
+        '[--middlware MIDDLEWARE_PATH]',
```
SounD120:
```suggestion
        '[--middleware MIDDLEWARE_PATH]',
```

RedMickey:
```suggestion
        '[--middleware MIDDLEWARE_PATH]',
```

PR #310, opened 2020-01-24
Allow custom plugins written in nodeJS

mfish33:
This is related to PR: https://github.com/microsoft/cordova-simulate/pull/308. Instead of handling custom node plugins through the https(in the diagram in the aforementioned PR), I propose a new type of plugin file that can be run in the Node environment. This allows for a wider range of plugins such as app cache stores and can allow for the integration of existing node modules, in plugin mocks. Further, this PR would remove my suggestion of having a default middleware.

The Order of exec calls now becomes:
1. App host handlers
2. Simulate_gap handlers
3. Sim host handlers


JiglioNero:
Hi @mfish33  Thanks a lot for your contributions!
We’re investigating your PR and for now, as we understand, this feature could be used in the following cases:
- To mock some existing `exec` request handlers in plugins
- To add `exec` request handlers to plugins that don’t provide them to enhance plugins’ functionality
- To create new plugins for cordova-simulate to handle different `exec` requests

Could you please tell us if we understood the idea of the feature correctly? 
Could you please also clarify a bit for what purpose "gap plugins" could be used in middleware?

While reviewing the PR we found out some points that could be improved:
- It seems that using `process.cwd()` [here](https://github.com/mfish33/cordova-simulate/blob/node-plugins/src/server/node-plugins.js) can lead to errors if cwd differs from a project directory, in case the `--dir` parameter is passed to the simulate command. `--dir` parameter points to the project directory, which may not coincide with `process.cwd()`
- We suppose that it would be better to add a parameter to activate "gap plugins" feature instead of looking for `simulate_gap.js` files by default

EzioLi01:
Close this PR since it has no updates for long time. Please reopen this if it still need to work.

PR #465, opened 2024-02-07
Add chromium path arguments

RodBerw:


bin/simulate
```diff
@@ -35,7 +35,7 @@ function unknownOptHandler(optName) {
 function processArgs() {
     try {
         var parserOptions = {
-            string: ['dir', 'simhostui', 'target', 'simulationpath', 'device', 'lang', 'theme', 'middleware'],
+            string: ['dir', 'simhostui', 'target', 'simulationpath', 'device', 'lang', 'theme', 'middleware', 'chromiumPath'],
```
EzioLi01, line 38: Using lowercase for argument.

src/browsers/browser.js
```diff
@@ -73,6 +74,7 @@ function getBrowser(target, dataDir, url) {
     const browsers = {
         win32: {
             chrome: `chrome --user-data-dir=%TEMP%\\${dataDir}`,
+            chromium: chromiumPath || 'chromium',
```
EzioLi01, line 77: Only path is enough for windows I think.

EzioLi01:
Generally I think chromium can only support windows here. You can add some logs:
1. In win32 but not path argument set, will throw exception
2. In other OSs and set path, show log to notice that not support.

bin/simulate
```diff
@@ -77,7 +77,8 @@ function processArgs() {
             middleware: parsed.middleware,
             generateids: parsed.generateids,
             spaurlrewrites: parsed.spaurlrewrites,
-            showbrowser: parsed.showbrowser
+            showbrowser: parsed.showbrowser,
+            chromiumPath: parsed.chromiumPath
```
EzioLi01, line 81: Also lowercase?

bin/simulate
```diff
@@ -102,7 +103,8 @@ function getUsageString() {
         '[--middleware MIDDLEWARE_PATH]',
         '[--generateids (GENERATEIDS)]',
         '[--spaurlrewrites (SPA_URL_REWRITES)]',
-        '[--showbrowser (SHOWBROWSER)]'
+        '[--showbrowser (SHOWBROWSER)]',
+        '[--chromiumPath CHROMIUM_PATH]'
```
EzioLi01, line 107: lowercase

bin/simulate
```diff
@@ -77,8 +77,10 @@ function processArgs() {
             middleware: parsed.middleware,
             generateids: parsed.generateids,
             spaurlrewrites: parsed.spaurlrewrites,
-            showbrowser: parsed.showbrowser
+            showbrowser: parsed.showbrowser,
+            chromiumpath: parsed.chromiumpath
         };
+        console.log(chromiumpath)
```
EzioLi01, line 83: Remove this log.

src/simulate.js
```diff
@@ -17,18 +17,27 @@ var simulate = function (opts) {
...
     var simulator = new Simulator(opts);
     var showBrowser = opts.showbrowser;
+    var chromiumPath = opts.chromiumpath;
 
     if (!showBrowser) {
         var noBrowserMessage = 'The argument `showbrowser` is set to false. Please load simulated application in browser manually if needed.';
         log.warning(noBrowserMessage);
     }
 
+    if(process.platform === 'win32' && !chromiumPath){
```
EzioLi01, line 27: We have several browsers as the target, only chromium needs this verification.

src/simulate.js
```diff
@@ -17,18 +17,27 @@ var simulate = function (opts) {
...
     var simulator = new Simulator(opts);
     var showBrowser = opts.showbrowser;
+    var chromiumPath = opts.chromiumpath;
 
     if (!showBrowser) {
         var noBrowserMessage = 'The argument `showbrowser` is set to false. Please load simulated application in browser manually if needed.';
         log.warning(noBrowserMessage);
     }
 
+    if(process.platform === 'win32' && target === 'Chromium' && !chromiumPath){
```
EzioLi01, line 27: Can we support both upper and lower case? Or only lower case.

src/simulate.js
```diff
@@ -17,18 +17,27 @@ var simulate = function (opts) {
...
     var simulator = new Simulator(opts);
     var showBrowser = opts.showbrowser;
+    var chromiumPath = opts.chromiumpath;
 
     if (!showBrowser) {
         var noBrowserMessage = 'The argument `showbrowser` is set to false. Please load simulated application in browser manually if needed.';
         log.warning(noBrowserMessage);
     }
 
+    if(process.platform === 'win32' && (target === 'chromium' || target === 'Chromium') && !chromiumPath){
```
EzioLi01, line 27: You can just use like `target = target.toLowerCase()` and verify.

RodBerw: I remember trying this at first, but it was not working for some reason at time, maybe I had altered the target var type while testing it before. I tryied it again and seems to work fine, thanks!

src/browsers/browser.js
```diff
@@ -62,7 +63,7 @@ function launchBrowser(opts) {
     }
 }
 
-function getBrowser(target, dataDir, url) {
+function getBrowser(target, dataDir, chromiumPath, url) {
```
EzioLi01, line 66: Please move `chromiumPath` to the last argument to avoid some unit test failures.

EzioLi01:
LGTM, thanks for the contribution!

